PR #88, opened 2013-10-08
Fix spaces encoding in parameters

Xowap:
OAuth requires spaces to be encoded as %20 instead of + in order to generate a valid signature.

Explained in http://troy.yort.com/2-common-problems-with-oauth-client-libraries/

Before this patch, I was unable to use Yahoo!'s BOSS Placefinder API, which now seems to be responding correctly to me, instead of what it used to say:

```
<?xml version='1.0' encoding='UTF-8'?>
<yahoo:error xmlns:yahoo='http://yahooapis.com/v1/base.rng'
  xml:lang='en-US'>
  <yahoo:description>Please provide valid credentials. OAuth oauth_problem="signature_invalid", realm="yahooapis.com"</yahoo:description>
</yahoo:error>
```

### DIFF
--- a/requests_oauthlib/oauth1_session.py
+++ b/requests_oauthlib/oauth1_session.py
@@ -7,6 +7,8 @@ except ImportError:
 
 from oauthlib.common import add_params_to_uri, urldecode
 from oauthlib.oauth1 import SIGNATURE_HMAC, SIGNATURE_TYPE_AUTH_HEADER
+from requests.utils import to_key_val_list
+from urllib import urlencode
 import requests
 
 from . import OAuth1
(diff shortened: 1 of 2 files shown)


laughinghan:
Hi, I'm running into the same problem (with the exact same Yahoo BOSS Placefinder API, fancy that), and I'm trying to understand the workarounds. This PR forks `requests_oauthlib`, [this other blogpost author](https://www.maxcdn.com/blog/rfc-3986-web-encoding-hell/) forked `requests`; what's your recommended fix "above the fold", so to speak? Right now I'm calling `urllib.urlencode(params).replace('+', '%20')` myself and passing that in as `params` to `requests`, because `requests` is awesome and will accept params as a `str` rather than `dict`, but it's a pain because `urllib` is Unicode-unaware and instead expects UTF-8 encoded `str()`s: http://stackoverflow.com/a/6481120/362030

singingwolfboy:
There's been no activity on this pull request for over two years, and it has conflicts with the master branch. As a result, I'm closing this pull request. If we determine that we still want this change after all, we can open a new pull request for it, based on the latest version of master.

PR #149, opened 2014-09-22
add POST params to fetch_request_token in oauth1_session

kasajei:
Some service need POST params for the method.


Lukasa:
Closed in favour of #226.

Issue #164, opened 2014-12-22
Test test_signature_methods fails when optional dependency not installed


koobs:
The following test fails when the optional (rsa) dependency pycrypto is not installed. 

It should skip instead, or be added to tests_require. Both would be best.

```
======================================================================
ERROR: test_signature_methods (tests.test_oauth1_session.OAuth1SessionTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/mock.py", line 1201, in patched
    return func(*args, **keywargs)
  File "/mnt/home/user/repos/freebsd/ports/www/py-requests-oauthlib/work/requests-oauthlib-0.4.2/tests/test_oauth1_session.py", line 102, in test_signature_methods
    auth.post('https://i.b')
  File "/mnt/home/user/repos/freebsd/ports/www/py-requests-oauthlib/work/requests-oauthlib-0.4.2/requests-2.5.0-py2.7.egg/requests/sessions.py", line 504, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/mnt/home/user/repos/freebsd/ports/www/py-requests-oauthlib/work/requests-oauthlib-0.4.2/requests-2.5.0-py2.7.egg/requests/sessions.py", line 447, in request
    prep = self.prepare_request(req)
  File "/mnt/home/user/repos/freebsd/ports/www/py-requests-oauthlib/work/requests-oauthlib-0.4.2/requests-2.5.0-py2.7.egg/requests/sessions.py", line 378, in prepare_request
    hooks=merge_hooks(request.hooks, self.hooks),
  File "/mnt/home/user/repos/freebsd/ports/www/py-requests-oauthlib/work/requests-oauthlib-0.4.2/requests-2.5.0-py2.7.egg/requests/models.py", line 308, in prepare
    self.prepare_auth(auth, url)
  File "/mnt/home/user/repos/freebsd/ports/www/py-requests-oauthlib/work/requests-oauthlib-0.4.2/requests-2.5.0-py2.7.egg/requests/models.py", line 492, in prepare_auth
    r = auth(self)
  File "/mnt/home/user/repos/freebsd/ports/www/py-requests-oauthlib/work/requests-oauthlib-0.4.2/requests_oauthlib/oauth1_auth.py", line 87, in __call__
    unicode(r.url), unicode(r.method), None, r.headers)
[...]

PR #167, opened 2015-01-29
Issue #164: pycrypto dependency

jayvdb:


singingwolfboy:
[oauthlib no longer depends on pycrypto at all.](https://github.com/idan/oauthlib/commit/fc2d1218e015adcf65ba01965884430c47895cd0)

Lukasa:
Clearly my totally forgetting about this issue has resolved the problem. Thanks @singingwolfboy!

jayvdb:
Resubmitted, amended due to switch from pycrypto to cryptography, as #196

PR #172, opened 2015-02-10
Check if client exists before referencing client_id

singingwolfboy:
I'm trying to instantiate an `OAuth2Session` object, and populate information (like `client_id`) *after* the initialization code has already run. This change is necessary for that -- otherwise, I get `AttributeError: 'NoneType' object has no attribute 'client_id'`

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -64,7 +64,9 @@ class OAuth2Session(requests.Session):
         :param kwargs: Arguments to pass to the Session constructor.
         """
         super(OAuth2Session, self).__init__(**kwargs)
-        self.client_id = client_id or client.client_id
+        self.client_id = client_id
+        if client is not None and not self.client_id:
+            self.client_id = client.client_id
         self.scope = scope
         self.redirect_uri = redirect_uri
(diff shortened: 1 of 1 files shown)


singingwolfboy:
ping

requests_oauthlib/oauth2_session.py
```diff
@@ -64,7 +64,9 @@ def __init__(self, client_id=None, client=None, auto_refresh_url=None,
         :param kwargs: Arguments to pass to the Session constructor.
         """
         super(OAuth2Session, self).__init__(**kwargs)
-        self.client_id = client_id or client.client_id
+        self.client_id = client_id
+        if client and not self.client_id:
```
Lukasa: Can this be an explicit `None` check? `if client is not None`

singingwolfboy: Fixed.

singingwolfboy:
ping

singingwolfboy:
@Lukasa Thanks! Any chance you can make another release on PyPI to bring this fix out to the world? (You might want to address #165 while you're at it.)

Lukasa:
I've just pushed 0.5.0. =)

PR #177, opened 2015-03-28
cleanup docstring of OAuth1Session to fix a typo/usage error

almartin82:
just a simple fix on it's vs its


Lukasa:
Thanks so much @almartin82! :cake: :sparkles: :cookie:

PR #180, opened 2015-05-10
Add Slack compliance

mikoim:
This PR fix a bug that its get method always returns 'not_authed' on Slack.

To call api methods on Slack, it require a URL including 'token' .


Lukasa:
LGTM! :cake: :space_invader: :sparkles:

PR #183, opened 2015-06-05
Update Legacy Application Flow documentation

CorwinTanner:
I added a basic description of the Legacy Application Flow. It's pretty simple, but it took a bit of tooling around so this should save some people the trouble.

I discovered that client_id is not passed from the OAuth2Session or LegacyApplicationClient when fetch_token is called, which I believe to be a bug.

### DIFF
--- a/docs/oauth2_workflow.rst
+++ b/docs/oauth2_workflow.rst
@@ -109,8 +109,25 @@ Documentation coming soon. Want to help? Why not `write this section`_?
 Legacy Application Flow
 -----------------------
 
-Documentation coming soon. Want to help? Why not `write this section`_?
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks for the docs patch @CorwinTanner! I'm going to merge this now, but can you elaborate on what you mean about this bug?

CorwinTanner:
An instance of LegacyApplicationClient is initialized with client_id and then passed into the initializer for OAuth2Session, which also stores a reference to client_id. However, when fetch_token is called on OAuth2Session, client_id is not included in the request. My workaround was to pass client_id as a kwarg in fetch_token, but it should be passed automatically since the session object already has it.

I imagine it's a very low priority bug, but a bug nonetheless. If I get time this weekend, I will propose a fix.

PR #195, opened 2015-09-18
When refreshing tokens, pass through **kwargs

jsfan:
When tokens are refreshed, the **kwargs are not passed through to refresh_tokens. This will lead to parameters like SSL verification, authentication and timeout settings to be dropped.

This patch fixes the problem.


Lukasa:
Seems reasonable to me, thanks! :cake:

PR #168, opened 2015-01-29
Travis testing of setup.py dependency management

jayvdb:
This fails due to issue #164, but I'll refresh it after that has landed and addressing any code review

PR #196, opened 2015-09-23
Skip rsa tests when dependencies not met

jayvdb:
Fixes issue #164

### DIFF
--- a/tests/test_oauth1_session.py
+++ b/tests/test_oauth1_session.py
@@ -1,4 +1,4 @@
-from __future__ import unicode_literals
+from __future__ import unicode_literals, print_function
 import mock
 import unittest
 import sys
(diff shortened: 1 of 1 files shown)


tests/test_oauth1_session.py
```diff
@@ -95,6 +109,11 @@ def fake_send(r, **kwargs):
     @mock.patch('oauthlib.oauth1.rfc5849.generate_timestamp')
     @mock.patch('oauthlib.oauth1.rfc5849.generate_nonce')
     def test_signature_methods(self, generate_nonce, generate_timestamp):
+        try:
+            import cryptography
+        except ImportError:
+            raise unittest.SkipTest('cryptography not available')
```
singingwolfboy: Rather than doing this four-line check, I'd suggest doing this at the top of the file:
```python
try:
    import cryptography
except ImportError:
    cryptography = None
```
and the using the [`skipUnless()`](https://docs.python.org/2/library/unittest.html#unittest.skipUnless) decorator on tests that require the `cryptography` module:
```python
@unittest.skipUnless(cryptography, "cryptography module is required")
def test_signature_methods(self, generate_nonce, generate_timestamp):
    # ...
```

jayvdb: I've moved the import at the top.
Unfortunately skipUnless isn't available on Python 2.6's `unittest`, and using monkey patching for that is going too far IMO.  We could switch to using `unittest2` on Python 2.6.

jayvdb:
Travis build with revised patch and the #168 patch: https://travis-ci.org/jayvdb/requests-oauthlib/builds/82236300

tests/test_oauth1_session.py
```diff
@@ -20,6 +25,20 @@
...
     bytes_type = str
 
+# Monkey patch Python 2.6 unittest
+if not hasattr(unittest, 'SkipTest'):
+    unittest.SkipTest = RuntimeWarning
+    unittest.TestResult.real_add_error = unittest.TestResult.addError
+
+    def patched_addError(self, test, exc_info):
+        if exc_info[0] is RuntimeWarning:
+            print(str(exc_info[1]), end=' ', file=sys.stderr)
```
Lukasa: Why are we bothering to print anything here?

jayvdb: So that the unittest output shows something useful. i.e.
https://travis-ci.org/jayvdb/requests-oauthlib/jobs/82236301#L176

jayvdb: Do you want me to add unittest2 support so that this workaround is needed?

Lukasa: That might be better, yeah.

Issue #178, opened 2015-03-29
Bad error message for wrong redirect URI


JensErat:
I just worked through the [Facebook authentication tutorial](https://requests-oauthlib.readthedocs.org/en/latest/examples/facebook.html), and was stuck for quite some time until I started tracing the logs.

## How to Reproduce

My problem was I registered a URI with trailing `/` (let it be `https://example.org/`) at facebook, but used the same URI without slash `https://example.org` in the python code.

`r.text` in `oauth2_session.py` contained a helpful error message:

> {"error":{"message":"Error validating verification code. Please make sure your redirect_uri is identical to the one you used in the OAuth dialog request","type":"OAuthException","code":100}}

## Expected Output

The error message is printed to the developer (eg., a reasonable exception is thrown).

## Actual Output

Instead it was hidden from me, and I got a barely helpful and misleading traceback instead:

    Traceback (most recent call last):
      File "./test.py", line 30, in <module>
        facebook.fetch_token(token_url, client_secret=client_secret, authorization_response=redirect_response)
      File "/usr/local/lib/python2.7/dist-packages/requests_oauthlib/oauth2_session.py", line 199, in fetch_token
        self._client.parse_request_body_response(r.text, scope=self.scope)
      File "/usr/lib/python2.7/dist-packages/oauthlib/oauth2/rfc6749/clients/web_application.py", line 271, in parse_request_body_response
        self.token = parse_token_response(body, scope=scope)
      File "/usr/lib/python2.7/dist-packages/oauthlib/oauth2/rfc6749/parameters.py", line 303, in parse_token_response
        validate_token_parameters(params, scope)
      File "/usr/lib/python2.7/dist-packages/oauthlib/oauth2/rfc6749/parameters.py", line 313, in validate_token_parameters
        raise MissingTokenError(description="Missing access token parameter.")
    oauthlib.oauth2.rfc6749.errors.MissingTokenError

Lukasa:
If you can still easily reproduce this, what's the status code of that response?

JensErat:
Which status response, where can I fetch it from? There is a code 100 in `r.text`, I guess that's not what you're after?

Lukasa:
That's not what I'm after, I'm after the HTTP status code. Where you print `r.text` you can also print `r.status_code`.

JensErat:
Aye, `r` for response. ;)

[...]

PR #198, opened 2015-10-21
Raise possible exceptions in refresh_token

amegianeg:
Hi!
My scenario: trying to refresh a token, Digital Ocean's API returned 429 (too many requests) but no error were raised, so the exception was masked as `MissingTokenError(description="Missing access token parameter.")`.


Lukasa:
@amegianeg Sorry, we've had a back-and-forth on this for quite a while. We added it in #178, then removed it in #197, and at the moment I don't plan to re-add it. You can in principle get meaningful content back in an error code, so I'm extremely reluctant to add a catch-all `raise` in this spot.

PR #200, opened 2015-10-22
Added note to Real World Example

### LINKED ISSUE #199: Documentation Problem in Real World Example
Hey,

in the [real world example](https://github.com/requests/requests-oauthlib/blob/master/docs/examples/real_world_example.rst) you are using Githubs API as a case study and Flask as the framework. It would probably be worth a mention that with Github's API, the `redirect_uri` parameter is optional, especially if you are using flask on the other side, as `flask-oauthlib` will require this parameter and if users want to create something based on that example, this will probably be a minor pitfall.

I can issue a PR to fix this if you agree that this is somewhat confusing.

Cheers

hellerve:
This closes #199. I am not sure whether the wording is clear (it sure is not as concise as I would like it to be). Any suggestions on how to rephrase are appreciated.

Cheers

### DIFF
--- a/docs/examples/real_world_example.rst
+++ b/docs/examples/real_world_example.rst
@@ -8,7 +8,7 @@ a web application might look like using the `Flask web framework
 <http://flask.pocoo.org/>`_ and GitHub as a provider. It should be easily
 transferrable to any web framework.
 
-.. note
+.. note::
   While the flow remains the same across msot providers, Github is special in that
   the `redirect_uri` parameter is optional. This means that it may be necessary to
   explicitly pass a redirect_uri to the `OAuth2Session` object (e.g. when creating
--- a/docs/examples/real_world_example.rst
+++ b/docs/examples/real_world_example.rst
@@ -8,6 +8,12 @@ a web application might look like using the `Flask web framework
 <http://flask.pocoo.org/>`_ and GitHub as a provider. It should be easily
 transferrable to any web framework.
 
+.. note
+  While the flow remains the same across msot providers, Github is special in that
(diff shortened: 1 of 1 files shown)


docs/examples/real_world_example.rst
```diff
@@ -8,6 +8,12 @@ a web application might look like using the `Flask web framework
 <http://flask.pocoo.org/>`_ and GitHub as a provider. It should be easily
 transferrable to any web framework.
 
+.. note::
+  While the flow remains the same across msot providers, Github is special in that
```
Lukasa: Can you bring this whole block up to the `.. note::` and then indent accordingly? That is:

```
.. note:: While the flow remains the same across most providers, GitHub
          is special in that...
```

Note also the spelling of `most`. =)

docs/examples/real_world_example.rst
```diff
@@ -8,6 +8,12 @@ a web application might look like using the `Flask web framework
...
+  the `redirect_uri` parameter is optional. This means that it may be necessary to
+  explicitly pass a redirect_uri to the `OAuth2Session` object (e.g. when creating
+  a custom OAuthProvider with `flask-oauthlib`).
```
Lukasa: For rst, you need to use double-backticks. That is:

    ``OAuth2Session``

Lukasa:
Some inline notes, but otherwise this is good. =D

hellerve:
As you may have seen, I have no idea of rst (mostly a Markdown user). I will do as you told me and then report back.

Lukasa:
Thanks! :sparkles: :cake: :sparkles:

PR #208, opened 2015-12-12
Getters/Setters for wrapped attributes

singingwolfboy:
Code that uses requests-oauthlib may want to modify values in the oauth2 session client, but requests-oauthlib doesn't expose them in a consistent way. By exposing getters and setters, other software and libraries that wrap requests-oauthlib can modify these values in a consistent way.

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -64,16 +64,15 @@ class OAuth2Session(requests.Session):
         :param kwargs: Arguments to pass to the Session constructor.
         """
         super(OAuth2Session, self).__init__(**kwargs)
+        self._client = client or WebApplicationClient(client_id, token=token)
+        self.token = token or {}
         self.scope = scope
         self.redirect_uri = redirect_uri
-        self.token = token or {}
         self.state = state or generate_token
         self._state = state
         self.auto_refresh_url = auto_refresh_url
         self.auto_refresh_kwargs = auto_refresh_kwargs or {}
         self.token_updater = token_updater
-        self._client = client or WebApplicationClient(client_id, token=token)
-        self._client._populate_attributes(token or {})
 
         # Allow customizations for non compliant providers through various
         # hooks to adjust requests and responses.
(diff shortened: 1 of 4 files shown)


requests_oauthlib/oauth2_session.py
```diff
@@ -64,19 +64,15 @@ def __init__(self, client_id=None, client=None, auto_refresh_url=None,
...
         self.scope = scope
         self.redirect_uri = redirect_uri
-        self.token = token or {}
         self.state = state or generate_token
         self._state = state
         self.auto_refresh_url = auto_refresh_url
         self.auto_refresh_kwargs = auto_refresh_kwargs or {}
         self.token_updater = token_updater
-        self._client = client or WebApplicationClient(client_id, token=token)
-        self._client._populate_attributes(token or {})
```
Lukasa: Why did we lose this function call?

singingwolfboy: Because this function call is in the setter for `self.token`, so it gets called automatically on line 68.

requests_oauthlib/oauth2_session.py
```diff
@@ -64,19 +64,15 @@ def __init__(self, client_id=None, client=None, auto_refresh_url=None,
         :param kwargs: Arguments to pass to the Session constructor.
         """
         super(OAuth2Session, self).__init__(**kwargs)
-        self.client_id = client_id
-        if client is not None and not self.client_id:
-            self.client_id = client.client_id
+        self._client = client or WebApplicationClient(client_id, token=token)
+        self.token = token or {}
```
Lukasa: This logic is not quite the same: falsy objects for `client` will now be not set. I'm nervous about this as a back-incompatible change.

singingwolfboy: I'm confused -- I don't see the backwards-incompatibility, myself. Can you provide a concrete example demonstrating the behavior that changes in this pull request? An actual automated test would be awesome, but even just a few lines of Python in a comment would be helpful.

Lukasa: Apologies, I was clearly very tired: you're right. =)

requests_oauthlib/oauth2_session.py
Lukasa: Aha, missed that: thanks!

Lukasa:
Thanks @singingwolfboy!

PR #209, opened 2015-12-13
Subclass requests.auth.AuthBase

singingwolfboy:
The official documentation says to [subclass `requests.auth.AuthBase`](http://docs.python-requests.org/en/latest/user/authentication/#new-forms-of-authentication), so requests-oauthlib should do that. Requests-oauthlib only supports requests at version 2.0.0 and above, and [the `AuthBase` class existed at version 2.0.0](https://github.com/kennethreitz/requests/blob/2.0/requests/auth.py#L34), so we don't need to worry about ImportErrors.


sigmavirus24:
Thanks @singingwolfboy

PR #214, opened 2015-12-20
Add refresh_token_request compliance hook

jimcortez:
This hook is to be able to modify refresh token requests. It is useful when doing automatic refreshes.

Also added example for the fitbit developer api, which requires a Authorization header at refresh time.


jimcortez:
This should address issue #128

Lukasa:
Do you really mean #128?

jimcortez:
@Lukasa yes, I did. :blush:

edited.

sigmavirus24:
This has merge conflicts that need to be resolved.

singingwolfboy:
Looks like the Fitbit compliance fix was merged in #261. As a result, I'm closing this pull request.

PR #215, opened 2015-12-23
Use requests-mock library, add tests for Slack compliance fix

singingwolfboy:
The tests are clearer and more accurate if we use the [requests-mock](https://github.com/openstack/requests-mock) library for mocking responses from requests. This pull request also adds a test for the Slack compliance fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ python:
   - 3.5
 install:
   - pip install -r requirements.txt
+  - pip install requests-mock
   - pip install coveralls
   - if [[ $TRAVIS_PYTHON_VERSION == "2.6" ]]; then pip install unittest2; fi
 script:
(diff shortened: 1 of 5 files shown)


sigmavirus24:
If we're going to hand write response data, I'd rather we use requests-mock. It's maintained and raises errors that are not "ConnectionError" when a request changes. responses is a poor excuse for a mocking library for requests.

Also, I'd rather we not mock any response bodies by hand, but that's a different arguement entirely.

singingwolfboy:
@sigmavirus24 I had no idea that responses was not maintained, and that requests-mock was the library to use instead. I've updated my pull request. Does it look good?

sigmavirus24:
:+1: from me.

Lukasa:
:+1: from me too.

singingwolfboy:
:dancers:

PR #219, opened 2016-01-12
Don't nullify token before refresh

dsanader:
Hello,

I'm using an OAuth2Session object, with the token autorefresh feature, inside a library. This involves some level of concurrent requests. 

I've noticed that some requests were issued without the Authorization header. This was happening around the refresh requests. I've tracked it to the fact that the token was nullified inside the refresh_token method, causing my OAuth2Session instance to lack a token for a short period of time (until a new one is fetched).

The attached changes fixed the issue for me (no token reset, and issue a regular requests.post instead of using the class equivalent). That way the token is simply replaced.

Regards.

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -269,9 +269,7 @@ class OAuth2Session(requests.Session):
         if not is_secure_transport(token_url):
             raise InsecureTransportError()
 
-        # Need to nullify token to prevent it from being added to the request
         refresh_token = refresh_token or self.token.get('refresh_token')
-        self.token = {}
 
         log.debug('Adding auto refresh key word arguments %s.',
                   self.auto_refresh_kwargs)
@@ -288,8 +286,8 @@ class OAuth2Session(requests.Session):
                 ),
             }
 
-        r = self.post(token_url, data=dict(urldecode(body)), auth=auth,
-                      timeout=timeout, headers=headers, verify=verify)
+        r = requests.post(token_url, data=dict(urldecode(body)), auth=auth,
+            timeout=timeout, headers=headers, verify=verify)
         log.debug('Request to refresh token completed with status %s.',
                   r.status_code)
         log.debug('Response headers were %s and content %s.',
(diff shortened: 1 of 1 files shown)


Lukasa:
Unfortunately, this fix isn't going to work.

Using a normal `post` instead of the Session one means that a whole lot of settings the user may have applied via Transport Adapters, or on the Session itself. While the fix may work specifically in your case, it is unlikely to work generally and a little surprising.

It may be more fruitful to add a flag that can be passed into `post` (and that ends up in `OAuth2Session.request`) that skips the `if self.token` logic in the `OAuth2Session.request` method. That should be fairly safe, and will ensure that the fix behaves properly in the general case.

Are you interested in making that fix?

dsanader:
I must confess I haven't given much thought to the bigger picture.

I think I've updated my branch accordingly. Is this what you meant ?

Lukasa:
Yup, that's a lot closer! The name isn't great though: let's call it `withhold_token`.

dsanader:
Flag renamed.

requests_oauthlib/oauth2_session.py
```diff
@@ -289,7 +287,7 @@ def refresh_token(self, token_url, refresh_token=None, body='', auth=None,
             }
 
         r = self.post(token_url, data=dict(urldecode(body)), auth=auth,
-                      timeout=timeout, headers=headers, verify=verify)
+            timeout=timeout, headers=headers, verify=verify, withhold_token=True)
```
Lukasa: Can we fix up the indenting here? You'll probably need to move withold_token to the next line to make everything fit an appropriate line length.

Lukasa:
Fantastic, thanks! One small style nit and we're good to go!

dsanader:
I've matched indenting to the way it looked in the fetch_token method (lines 212 and 218).

Should I fix these lines too ?

Lukasa:
@dsanader Ah, good spot. Let's leave it then.

Thanks! :sparkles: :cake: :sparkles:

dsanader:
Cool! Thanks for you quick and kind review.

Cheers.

PR #226, opened 2016-02-08
Allow kwargs for fetch_request_token and fetch_access_token to be passed to requests.session

Christian-E:
This was basically suggested in the comments of #149. As i can't commit to this open pull request i make a new one. 
Allow users to pass kwargs to fetch_request_token and fetch_access_token that are then passed to the post method of requests.session. 
I need this for disabling SSL-verification for certain calls. But I don't think it makes sense to implement every possible parameter individually.


Lukasa:
This looks good to me, thank you! :sparkles: :cake: :sparkles:

Issue #227, opened 2016-02-24
InvalidClient: HTTPBasicAuth with username and password for Legacy Applications


synasius:
I stumbled upon this error while trying to use requests-oauthlib with an oauth2 server created using django-oauth-toolkit. I think that there is a misunderstanding of section 2.3.1 of the RFC 6749 (here is the link https://tools.ietf.org/html/rfc6749#section-2.3.1) that is also the cause of #221 and #218 

The RFC says:
> Clients in possession of a client password MAY use the HTTP Basic
   authentication scheme as defined in [RFC2617] to authenticate with
   the authorization server.

The `client password` is the `client secret`. How should we create a Basic Auth Header? the RFC says:
> The *client identifier* is encoded using the
   "application/x-www-form-urlencoded" encoding algorithm per
   Appendix B, and the encoded value is used as the username; the *client
   password* is encoded using the same algorithm and used as the
   password.

client_id is used as username.
client_secret is used as password.

So I think that the header should be create as follows (and that is what we expect in django-oauth-toolkit):
```
auth = requests.auth.HTTPBasicAuth(client_id, client_secret)
```
the problem is that `username` and `password` are used instead, which are user's credentials and not the application's credentials.

Also, the RFC example shows the following header
> Authorization: Basic czZCaGRSa3F0Mzo3RmpmcDBaQnIxS3REUmJuZlZkbUl3

If you decode the base64 value you'll get:
```
>>> base64.b64decode('czZCaGRSa3F0Mzo3RmpmcDBaQnIxS3REUmJuZlZkbUl3')
>>> b's6BhdRkqt3:7Fjfp0ZBr1KtDRbnfVdmIw'
```
which are the client_id and client_secret  used in the example at the end of section 2.3.1

synasius:
Forgot to mention that, if you agree with my analysis, I can submit a PR.

Lukasa:
Reading the RFC appears to support your position, so I'm prepared to go with it. Feel free to submit a PR.

poswald:
Added a PR as a proposed fix. Please let me know if you think it will cause issues for other grant types or if the logic is not ideal in any way.

poswald:
This one can probably be closed too now that the PR is merged.

PR #229, opened 2016-03-14
Use client_id and client_secret as basic auth credentials if provided

poswald:
This should fix issue #227 

Open to review and suggestions if this will conflict with other grant flows.

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -197,10 +197,17 @@ class OAuth2Session(requests.Session):
                 redirect_uri=self.redirect_uri, username=username,
                 password=password, **kwargs)
 
-        if (not auth) and username:
-            if password is None:
-                raise ValueError('Username was supplied, but not password.')
-            auth = requests.auth.HTTPBasicAuth(username, password)
+        client_id = kwargs.get('client_id', '')
+        client_secret = kwargs.get('client_secret', '')
+        if not auth:
+            if client_id:
+                log.debug('Encoding client_id, client_secret as Basic auth credentials.')
+                auth = requests.auth.HTTPBasicAuth(client_id, client_secret)
+            elif username:
+                if password is None:
+                    raise ValueError('Username was supplied, but not password.')
+                log.debug('Encoding username, password as Basic auth credentials.')
+                auth = requests.auth.HTTPBasicAuth(username, password)
 
         headers = headers or {
             'Accept': 'application/json',
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -198,16 +198,11 @@ class OAuth2Session(requests.Session):
                 password=password, **kwargs)
 
         client_id = kwargs.get('client_id', '')
-        client_secret = kwargs.get('client_secret', '')
-        if not auth:
-            if client_id:
-                log.debug('Encoding client_id, client_secret as Basic auth credentials.')
-                auth = requests.auth.HTTPBasicAuth(client_id, client_secret)
-            elif username:
-                if password is None:
-                    raise ValueError('Username was supplied, but not password.')
-                log.debug('Encoding username, password as Basic auth credentials.')
-                auth = requests.auth.HTTPBasicAuth(username, password)
+        if client_id and not auth:
+            client_secret = kwargs.get('client_secret', '')
+            client_secret = client_secret if client_secret is not None else ''
+            log.debug('Encoding client_id "%s" with client_secret as Basic auth credentials.', client_id)
+            auth = requests.auth.HTTPBasicAuth(client_id, client_secret)
 
         headers = headers or {
             'Accept': 'application/json',
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -198,11 +198,17 @@ class OAuth2Session(requests.Session):
                 password=password, **kwargs)
 
         client_id = kwargs.get('client_id', '')
-        if client_id and not auth:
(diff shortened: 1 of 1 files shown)


Lukasa:
I think I'm happy with this change. @synasius, thoughts?

synasius:
@poswald @Lukasa 
I think you can remove the part where username and password are encoded as basic auth credentials. I don't know if this code is used elsewhere but there is no mention about user credentials being encoded in basic auth in the RFC6749.

Also it is worth checking if `client_secret` is empty before encoding `client_id` and `client_secret`.

``` python
if (not auth) and client_id:
    if not client_secret:
        raise ValueError(...)
    auth = requests.auth.HTTPBasicAuth(client_id, client_secret)
```

what do you think?

Lukasa:
Agreed that we should check for the empty `client_secret`. I'm less sure about removing the other blog: it *did* fix some users' problems, and I'm reluctant to remove it unconditionally.

synasius:
@Lukasa makes sense! :+1:

Lukasa:
@poswald Want to add that check for empty `client_secret`?

poswald:
OK, I didn't realize that username/password are never used as the basic auth header. I'll look into making that change. As for the `client_id` I may need some more explanation. You'll see in the [original issue I had raised on DOT](https://github.com/evonove/django-oauth-toolkit/issues/353) that was what I was trying to get working.

So when using the password grant type (Resource Owner Password Credentials Grant) and a public client, the `client_secret` isn't sent along. In this grant type you're authenticating the client and the user separately. If the client is a public client, you really can't make a very strong assertion that the request is authentic and the assumption is that the user should enter their credentials into the app because they trust it (for example it was downloaded from your website or the iOS app store under your name).

The [wording is a bit odd](https://tools.ietf.org/html/rfc6749#section-2.3.1) in that it's required but only if it's not blank:

    client_secret
         REQUIRED.  The client secret.  The client MAY omit the
         parameter if the client secret is an empty string.

It also is allowed to be blank in the DOT `Application` object. I was planning on opening a DOT issue regarding this because there's some assumptions built into it that kind of make it not work very cleanly. This issue came up for me while testing some code related to that.

Lukasa:
Ugh I hate the OAuth specification. Yes, you're right, it's not required. In that case, we want a defensive `client_secret = client_secret if client_secret is not None else ''`.

poswald:
How are we looking now?

Lukasa:
@poswald Sorry, I think the summary of the above conversation is that we wanted to keep the fallback to the username:password case: it clearly helped some users.

poswald:
Gah, sorry I missed your comment there. OK, let me look at the spec again too while I'm at it. You're certainly not alone in thinking the spec is annoying.

Lukasa:
@poswald Heh, yeah, working with OAuth2 is always "fun". Thanks for putting in so much effort though, it's really appreciated! :custard:

poswald:
So yeah, while it does look like authenticating user credentials in basic auth is not technically in the spec, who knows what servers are doing in the wild for people. This version fixes the issue without changing the behavior too much for people who weren't passing `client_id` in. There is a chance this could cause breakage for some people who used this implementation as a reference for what to do on the server side instead of going to the spec but I also think it fixes an issue where the client was not following the spec for the password grant type so it's good to do.

requests_oauthlib/oauth2_session.py
```diff
@@ -197,10 +197,18 @@ def fetch_token(self, token_url, code=None, authorization_response=None,
...
-        if (not auth) and username:
-            if password is None:
-                raise ValueError('Username was supplied, but not password.')
-            auth = requests.auth.HTTPBasicAuth(username, password)
+        client_id = kwargs.get('client_id', '')
+        if not auth:
+            if client_id:
+                log.debug('Encoding client_id "%s" with client_secret as Basic auth credentials.', client_id)
+                client_secret = kwargs.get('client_secret', '')
+                client_secret = client_secret if client_secret is not None else ''
```
Lukasa: This line is actually superfluous with the `.get` call above, and can be removed.

Lukasa:
> There is a chance this could cause breakage for some people who used this implementation as a reference for what to do on the server side instead of going to the spec but I also think it fixes an issue where the client was not following the spec for the password grant type so it's good to do.

Oh god I hope no-one implements their server with reference to this codebase, there are so many hacks to get around non-standard behaviours that the client codebase doesn't make it *at all clear* what a server is supposed to do!

One small note, and then I'll pull the trigger on this. =)

requests_oauthlib/oauth2_session.py
poswald: Strictly speaking I think it's not if your goal is to protect from passing `None` to the basic auth function. The `get` with a default is the default used if the key is missing. If someone calls `fetch_token( ... client_secret=None, ... )` then after `client_secret = kwargs.get('client_secret', '')` you'll find that `client_secret` is `None`:

```
>>> client_secret = {'client_secret': None}.get('client_secret', '')
>>> print client_secret
None
>>> client_secret = {}.get('client_secret', '')
>>> print client_secret
>>>```

Unless I'm mistaken about `None` not being possible or not needing to be protected against I think this is best. The other way I was considering was:

```
client_secret = {'client_secret': None}.get('client_secret', '') or ''
```

What do you think?

Lukasa: Mmm, yeah, so I was running under the assumption that `client_secret` wouldn't be passed as `None`, but now that you raise the possibility it occurs to me that I think some calls may default it that way. Let's leave this code in as defense-in-depth.

Lukasa:
Thanks @poswald! :sparkles: :cake: :sparkles:

PR #235, opened 2016-05-11
Allows tokens to be refreshed via a proxy

cmoman:
Hi.
I was using your library to access a webbased API via a proxy server.  The proxy requires authentication.  I found I couldn't refresh the token without adding the proxies parameter to the QAuth2Session class.
I have no idea if this raises other security issues but it works for my useage.  I thought I'd see if was useful in the general sense by creating this pull request.
Thanks,
Chris

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -258,7 +258,7 @@ class OAuth2Session(requests.Session):
         return self.token
 
     def refresh_token(self, token_url, refresh_token=None, body='', auth=None,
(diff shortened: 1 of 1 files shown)


Lukasa:
I think this change is entirely reasonable. =) Of course, you could have just set the proxies on the session itself (`sess.proxies = proxy_dict`), but there's nothing wrong with doing this as well.

cmoman:
Thanks for the comment about setting the session proxy. I hadn't thought to do this.

PR #236, opened 2016-05-13
Update google.rst

steven1227:
The url for token and authorization misplaced.


steven1227:
The URL for Google Oauth Authenticate and Google Oauth Token should swap into right place.

Lukasa:
Thanks! :sparkles:

steven1227:
@Lukasa Can I ask that is any way to revoke the token in this library?

Lukasa:
Not using this library, no.

PR #237, opened 2016-05-17
Allow explicit bypass of basic auth logic branch and fetching of tokens via proxy

malefice:
Some apps do not require the Basic Auth header when fetching a token, and some even outright fail. This happened to me with one of my projects earlier today. With this PR, the basic auth logic branch will only run if `auth` is explicitly `None` e.g. not passing the `auth` kwarg. Specifying `auth=False` will bypass said branch.

I also added support for running `fetch_token()` via a proxy since it is kind of weird that #235 exists for `refresh_token()`, but not for fetching. That and debugging is so much easier with an intercepting proxy which also helped me with the former problem.

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -198,7 +198,7 @@ class OAuth2Session(requests.Session):
                 password=password, **kwargs)
 
         client_id = kwargs.get('client_id', '')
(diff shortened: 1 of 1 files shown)


Lukasa:
I think this is reasonable enough. Thanks for the work! :sparkles:

PR #259, opened 2016-12-16
Fix `oauth_callback` change error

qnub:
Got error `request() got an unexpected keyword argument 'oauth_callback'` when try to call `.fetch_request_token(url=self.request_token_url, oauth_callback=self.redirect_uri)`

### DIFF
--- a/requests_oauthlib/oauth1_session.py
+++ b/requests_oauthlib/oauth1_session.py
@@ -344,7 +344,7 @@ class OAuth1Session(requests.Session):
 
     def _fetch_token(self, url, **request_kwargs):
         log.debug('Fetching token from %s using client %s', url, self._client.client)
-        r = self.post(url, **request_kwargs)
+        r = self.post(url, data=request_kwargs)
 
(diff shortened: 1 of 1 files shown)


Lukasa:
This change is definitely not right. `oauth_callback` is not a parameter we accept anywhere. What are you trying to achieve?

qnub:
@Lukasa sad, because is only way to force twitter to direct user back to URL other than set on apps.twitter.com.

Check «Step 1» here https://dev.twitter.com/web/sign-in/implementing:

> The only unique parameter in this request is `oauth_callback`, which must be a URL-encoded version of the URL you wish your user to be redirected to when they complete step 2.

Lukasa:
If you want to set a parameter, you should use `params=` instead. =)

qnub:
Have you tryed to do that? Seems it's not work.

Lukasa:
When you say it doesn't work, what specifically do you mean? Additionally, have you tried setting `data=` instead of `params=`?

PR #261, opened 2017-01-24
add a fitbit compliance fix

brad:
This adds a compliance fix for the Fitbit API so that errors that occur during access token or refresh token fetching get properly surfaced.
https://dev.fitbit.com/docs/oauth2/#authorization-errors

We have been using this successfully with the python-fitbit library for a few weeks now: https://github.com/orcasgit/python-fitbit/blob/master/fitbit/compliance.py


brad:
Thanks @Lukasa!

PR #263, opened 2017-02-16
Make this compatible with plentymarkets

bloodywing:
Plentymarkets uses CamelCase for token responses

See here: https://developers.plentymarkets.com/rest-doc/introduction

The Token Responses are from their docs, they are not mine.

### DIFF
--- a/requests_oauthlib/compliance_fixes/__init__.py
+++ b/requests_oauthlib/compliance_fixes/__init__.py
@@ -6,3 +6,4 @@ from .linkedin import linkedin_compliance_fix
 from .slack import slack_compliance_fix
 from .mailchimp import mailchimp_compliance_fix
 from .weibo import weibo_compliance_fix
+from .plentymarkets import  plentymarkets_compliance_fix
--- a/requests_oauthlib/compliance_fixes/__init__.py
+++ b/requests_oauthlib/compliance_fixes/__init__.py
@@ -6,4 +6,4 @@ from .linkedin import linkedin_compliance_fix
 from .slack import slack_compliance_fix
 from .mailchimp import mailchimp_compliance_fix
 from .weibo import weibo_compliance_fix
-from .plentymarkets import  plentymarkets_compliance_fix
+from .plentymarkets import plentymarkets_compliance_fix
--- a/requests_oauthlib/compliance_fixes/plentymarkets.py
+++ b/requests_oauthlib/compliance_fixes/plentymarkets.py
@@ -13,22 +13,23 @@ def plentymarkets_compliance_fix(session):
         else:
             return r
 
+        fixed_token = {}
+
         expires = token.pop('expiresIn')
         if expires is not None:
-            token['expires_in'] = expires
+            fixed_token['expires_in'] = expires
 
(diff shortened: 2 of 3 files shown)


requests_oauthlib/compliance_fixes/plentymarkets.py
```diff
@@ -0,0 +1,35 @@
...
+        access_token = token.pop('accessToken')
+        if access_token is not None:
+            token['access_token'] = access_token
+
+        refresh_token = token.pop('refreshToken')
+        if refresh_token is not None:
+            token['refresh_token'] = refresh_token
+
+        token['token_type'] = 'Bearer'
+        del token['tokenType']
```
Lukasa: Is there any reason not to just build a new dictionary doing this mapping? I'm nervous about this fix being a bit brittle in the face of changing dictionaries.

bloodywing: Oh no problem I rewrite it.
I copied the facebook fix and rewrite it for plenty, I just wanted to make it work.

requests_oauthlib/compliance_fixes/plentymarkets.py
```diff
@@ -0,0 +1,36 @@
...
+
+        access_token = token.pop('accessToken')
+        if access_token is not None:
+            fixed_token['access_token'] = access_token
+
+        refresh_token = token.pop('refreshToken')
+        if refresh_token is not None:
+            fixed_token['refresh_token'] = refresh_token
+
+        fixed_token['token_type'] = 'Bearer'
```
Lukasa: Sorry, I should have been clearer.

What I was going for here was more like:

```python
fixed_token = {}
for k, v in token.items():
    fixed_token[to_snake_case(k)] = v
```

This should resolve the problem in its entirety, and also defends against this changing in the future. Does that make more sense?

bloodywing: Gotcha.

requests_oauthlib/compliance_fixes/plentymarkets.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def plentymarkets_compliance_fix(session):
+
+    def _to_snake_case(n):
+        return re.sub('(.)([A-Z][a-z]+)', r'\1_\2', n).lower()
+
+    def _compliance_fix(r):
+
+        # Plenty returns the Token in CamelCase instead with _
```
Lukasa: `s/instead with/instead of with`

requests_oauthlib/compliance_fixes/__init__.py
```diff
@@ -6,3 +6,4 @@
 from .slack import slack_compliance_fix
 from .mailchimp import mailchimp_compliance_fix
 from .weibo import weibo_compliance_fix
+from .plentymarkets import  plentymarkets_compliance_fix
```
Lukasa: Ah, sorry I missed this: can we get a newline at the end here?

bloodywing: there are also 2 spaces. Cleaning this up.

PR #265, opened 2017-03-03
Modify token auto refresh handling.

btimby:
I think these changes make things more explicit. I chose to preserve old functionality with warnings rather than outright remove them.

All of the old functionality can be achieved through the use of `auto_refresh_kwargs`, the preferred explicit method.

I have not yet tested this in my codebase, but the old and new tests all pass. I added some missing tests for old functionality as well as tests for the new stuff.

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -1,13 +1,31 @@
 from __future__ import unicode_literals
 
 import logging
+import warnings
 
 from oauthlib.common import generate_token, urldecode
 from oauthlib.oauth2 import WebApplicationClient, InsecureTransportError
 from oauthlib.oauth2 import TokenExpiredError, is_secure_transport
+
 import requests
 
 log = logging.getLogger(__name__)
+
+
+# Preserve so we can call it.
+old_showwarning = warnings.showwarning
+
+def logwarning(message, category, filename, lineno, file=None):
+    """
+    Log warnings as well as print them.
+    """
+    log.warning(
+        '%s:%s: %s:%s' % 
+        (filename, lineno, category.__name__, message))
+    old_showwarning(message, category, filename, lineno, file=file)
+
+# Install our new handler.
+warnings.showwarning = logwarning
 
 
 class TokenUpdated(Warning):
@@ -289,7 +307,7 @@ class OAuth2Session(requests.Session):
                 log.debug('Encoding client_id "%s" with client_secret as Basic auth credentials.', client_id)
                 client_secret = kwargs.get('client_secret', '')
                 client_secret = client_secret if client_secret is not None else ''
-                log.warning('client_id provided, but client_secret missing')
+                warnings.warn('client_id provided, but client_secret missing')
                 auth = requests.auth.HTTPBasicAuth(client_id, client_secret)
 
         if headers is None:
@@ -325,7 +343,7 @@ class OAuth2Session(requests.Session):
             raise InsecureTransportError()
 
         if client_id or client_secret:
-            log.warning(
+            warnings.warn(
                 'Specify token refresh authentication in auto_refresh_kwargs.')
 
         if self.token and not withhold_token:
@@ -349,7 +367,7 @@ class OAuth2Session(requests.Session):
 
                     auth = kwargs.pop('auth', None)
                     if auth:
-                        log.warning('Specify token refresh authentication in '
+                        warnings.warn('Specify token refresh authentication in '
                                     'auto_refresh_kwargs.')
                         refresh_kwargs['auth'] = auth
 
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -279,9 +279,6 @@ class OAuth2Session(requests.Session):
 
         refresh_token = refresh_token or self.token.get('refresh_token')
 
-        log.debug('Adding auto refresh key word arguments %s.',
-                  self.auto_refresh_kwargs)
-        kwargs.update(self.auto_refresh_kwargs)
         body = self._client.prepare_refresh_body(body=body,
                 refresh_token=refresh_token, scope=self.scope, **kwargs)
         log.debug('Prepared refresh token request body %s', body)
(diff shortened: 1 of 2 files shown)


requests_oauthlib/oauth2_session.py
```diff
@@ -1,15 +1,33 @@
...
+    """
+    Log warnings as well as print them.
+    """
+    log.warning(
+        '%s:%s: %s:%s' % 
+        (filename, lineno, category.__name__, message))
+    old_showwarning(message, category, filename, lineno, file=file)
+
+# Install our new handler.
+warnings.showwarning = logwarning
```
Lukasa: I am very nervous about monkeypatching the warnings module. I'd rather not do it.

requests_oauthlib/oauth2_session.py
```diff
@@ -279,9 +297,7 @@ def refresh_token(self, token_url, refresh_token=None, body='', auth=None,
 
         refresh_token = refresh_token or self.token.get('refresh_token')
 
-        log.debug('Adding auto refresh key word arguments %s.',
-                  self.auto_refresh_kwargs)
-        kwargs.update(self.auto_refresh_kwargs)
+        log.debug('Kwargs: %s', kwargs)
```
Lukasa: This log message isn't really useful anymore: it might be better to remove it rather than leave it like this.

requests_oauthlib/oauth2_session.py
```diff
@@ -317,6 +333,7 @@ def request(self, method, url, data=None, headers=None, withhold_token=False,
         """Intercept all requests and add the OAuth 2 token if present."""
         if not is_secure_transport(url):
             raise InsecureTransportError()
+
```
Lukasa: Let's not add random changes. 😊

requests_oauthlib/oauth2_session.py
```diff
@@ -334,13 +351,42 @@ def request(self, method, url, data=None, headers=None, withhold_token=False,
...
-                    if client_id and client_secret and (auth is None):
+                    # Someday this code can be eliminated, we should be able to
+                    # simply pass **self.auto_refresh_kwargs to
+                    # `refresh_token()` what follows is to maintain
+                    # compatibility
+
+                    # Start with kwargs explicitly requested for refresh.
+                    refresh_kwargs = self.auto_refresh_kwargs.copy()
+
+                    if 'auth' in kwargs:
```
Lukasa: We should defend against `auth` being present but `None`.

requests_oauthlib/oauth2_session.py
```diff
@@ -334,13 +351,42 @@ def request(self, method, url, data=None, headers=None, withhold_token=False,
...
+                        warnings.warn('client_id argument supplied, Please '
+                                      'specify token refresh authentication in'
+                                      'auto_refresh_kwargs')
                         log.debug('Encoding client_id "%s" with client_secret as Basic auth credentials.', client_id)
-                        auth = requests.auth.HTTPBasicAuth(client_id, client_secret)
+                        if not client_secret:
+                            warnings.warn('client_id provided, but '
+                                          'client_secret missing')
+                            client_secret = ''
+                        refresh_kwargs['auth'] = \
```
Lukasa: A style note: we'd rather use implicit newlines inside parentheses than explicit ones.

requests_oauthlib/oauth2_session.py
```diff
@@ -334,13 +351,42 @@ def request(self, method, url, data=None, headers=None, withhold_token=False,
...
+                        # HTTP Basic auth header. It is preferred if the user
+                        # instead place an auth param into `auto_refresh_kwargs`
+                        warnings.warn('client_id argument supplied, Please '
+                                      'specify token refresh authentication in'
+                                      'auto_refresh_kwargs')
                         log.debug('Encoding client_id "%s" with client_secret as Basic auth credentials.', client_id)
-                        auth = requests.auth.HTTPBasicAuth(client_id, client_secret)
+                        if not client_secret:
+                            warnings.warn('client_id provided, but '
+                                          'client_secret missing')
```
Lukasa: Do we need this warning? I'm not certain how likely this is to be an error.

requests_oauthlib/oauth2_session.py
btimby: https://docs.python.org/2/library/warnings.html#warnings.showwarning

requests_oauthlib/oauth2_session.py
btimby: Yeah, that is a leftover, will remove it.

requests_oauthlib/oauth2_session.py
btimby: The newline does not show up in the resulting string, it is strictly for formatting, the two strings end up concatenated into one (single line) string.

requests_oauthlib/oauth2_session.py
btimby: I added it because the code replaces the missing argument with an empty string and continues with authentication. I thought that seemed like an omission on the callers part, something they might need warned about. This could be an incorrect assumption.

requests_oauthlib/oauth2_session.py
btimby: Yes, will update.

requests_oauthlib/oauth2_session.py
btimby: OK.

btimby:
Even after I make the requested changes, I am still testing, so please don't merge this. I will likely add new unit tests.

requests_oauthlib/oauth2_session.py
btimby: FYI... it was not "random" I was trying to make it more readable... Nevertheless the whitespace is removed.

requests_oauthlib/oauth2_session.py
btimby: Example:

```python
>>> print('this is a multi-'
... 'line string in the code but'
... ' a single line when printed.')
this is a multi-line string in the code but a single line when printed.
```

requests_oauthlib/oauth2_session.py
Lukasa: Sure, but it is still unnecessary for us to do this. We can just use a regular function, rather than change global state.

requests_oauthlib/oauth2_session.py
Lukasa: I know.  I'm saying don't use `\` (the explicit line continuation character), use parentheses instead.

requests_oauthlib/oauth2_session.py
Lukasa: Rather than do that, let's just require both the client ID and secret.

requests_oauthlib/oauth2_session.py
Lukasa: Legibility changes are fine, but they're better in seaparate PRs so that I can keep track of what is going on

benh57:
This change worked perfectly for me.

requests_oauthlib/oauth2_session.py
```diff
@@ -334,13 +349,42 @@ def request(self, method, url, data=None, headers=None, withhold_token=False,
...
-                    # We mustn't pass auth twice.
-                    auth = kwargs.pop('auth', None)
-                    if client_id and client_secret and (auth is None):
+                    # Someday this code can be eliminated, we should be able to
+                    # simply pass **self.auto_refresh_kwargs to
+                    # `refresh_token()` what follows is to maintain
+                    # compatibility
+
+                    # Start with kwargs explicitly requested for refresh.
+                    refresh_kwargs = self.auto_refresh_kwargs.copy()
```
east825: I wonder, isn't it going to change the semantic of `auto_refresh_kwargs` parameter? Previously it was intended to contain only key-value pairs to include into the body of the request, but now it also holds arguments of `refresh_token()` method such as `auth`, `timeout`, `proxies`, etc. If somebody wants to send `auth` parameter in the request body together with `refresh_token` and `grant_type`, this change will break their code.

singingwolfboy:
> Even after I make the requested changes, I am still testing, so please don't merge this. I will likely add new unit tests.

@btimby, are you still working on this? If you're still interested in getting this change merged, you should probably rebase it on the latest master, and then we can try again to get it reviewed and merged.

PR #266, opened 2017-03-08
add 'use_auto_state' parameter for 'authorization_url' method

MilovanovM:
'state' is an optional parameter for OAuth2

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -136,16 +136,16 @@ class OAuth2Session(requests.Session):
         """
         return bool(self.access_token)
 
-    def authorization_url(self, url, state=None, **kwargs):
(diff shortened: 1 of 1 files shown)


Lukasa:
So, while it's optional...do you really have to remove it? It's extremely important, and removing it exposes web services to CSRF attacks.

MilovanovM:
I completely agree, it's important, but I think it should be set by dev, not by library itself. The library should be flexible for all needs.
However, all old (and new) code will work as it did before because `use_auto_state` is `True` by default

Lukasa:
I guess what I'm trying to ask is: is there ever a sensible use-case where you'd want the state removed?

MilovanovM:
There is a specific use-case in our app. I think we could bypass it in another way.
Thanks, closing the PR out

PR #267, opened 2017-03-11
Added example and documentation edit for BackendApplicationClient

brockpalen:
The BackendApplicationClient example in the documentation fails for oauth2 gateways that expect the auth to only be in the header and not in the POST body.  In this example WSO2 as implemented by our org.  

The example was parted together from resolved bugs and appears to work against our gateway.  This request it to help others in the same situation.

### DIFF
--- a/docs/examples/wso2.rst
+++ b/docs/examples/wso2.rst
@@ -1,0 +1,31 @@
+WSO2 OAuth 2 Tutorial
+==========================
+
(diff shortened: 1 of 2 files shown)


Lukasa:
Thanks @brockpalen! I cleaned up some of the language and the formatting, but otherwise this looks really good! I'll wait until the build passes and then merge.

PR #271, opened 2017-04-17
Fix bytes/str mixups

jdufresne:
Fixes warnigns when running tests with all warnings on:

$ python -b -Wall -m nose

- DeprecationWarning: Please use assertEqual instead.
- BytesWarning: Comparison between bytes and string

### DIFF
--- a/tests/test_oauth2_session.py
+++ b/tests/test_oauth2_session.py
@@ -55,9 +55,7 @@ class OAuth2SessionTest(TestCase):
         token = 'Bearer ' + self.token['access_token']
 
         def verifier(r, **kwargs):
-            auth_header = r.headers.get('Authorization', None)
-            if 'Authorization'.encode('utf-8') in r.headers:
-                auth_header = r.headers['Authorization'.encode('utf-8')]
+            auth_header = r.headers.get(str('Authorization'), None)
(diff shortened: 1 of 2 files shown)


tests/test_oauth2_session.py
```diff
@@ -56,8 +56,6 @@ def test_add_token(self):
 
         def verifier(r, **kwargs):
             auth_header = r.headers.get('Authorization', None)
-            if 'Authorization'.encode('utf-8') in r.headers:
-                auth_header = r.headers['Authorization'.encode('utf-8')]
```
Lukasa: Why is this gone?

jdufresne:
> Why is this gone?

I have reexamined this bit of code. In all cases, the keys are the native `str` type. Not always unicode or always bytes. So I have updated to take this into account.

It was originally removed because, on Python3, it compared a `bytes` object to a `str` object with the `in` operator. Comparing these two types is usually a code smell as the code should _expect_ one or the other. When the `-b` flag is passed to `python`, it results in the warning "BytesWarning: Comparison between bytes and string", the change fixes this.

Lukasa:
Yup, that looks right. Ok, I'm happy. Thanks!

PR #277, opened 2017-05-02
Don't include kwargs of request() in the body of auto refresh request

east825:
Extra key-value pairs of the request body should be supplied using
the dedicated `auto_refresh_kwargs` attribute of the session object.

Meanwhile, several common parameters of request() and refresh_token()
are propagated to the latter, since, otherwise, auto refresh request
might end up without some critical connection parameters such as proxy
mapping.

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -339,8 +339,10 @@ class OAuth2Session(requests.Session):
                     if client_id and client_secret and (auth is None):
                         log.debug('Encoding client_id "%s" with client_secret as Basic auth credentials.', client_id)
                         auth = requests.auth.HTTPBasicAuth(client_id, client_secret)
+                    refresh_kwargs = {name: kwargs[name] for name in kwargs
+                                      if name in ('timeout', 'verify', 'proxies')}
                     token = self.refresh_token(
-                        self.auto_refresh_url, auth=auth, **kwargs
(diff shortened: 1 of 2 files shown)


Lukasa:
Before we go any further, is there a reason to prefer this PR to that in #265?

east825:
@Lukasa Yes, it addresses the same issue, but I have some [doubts](https://github.com/requests/requests-oauthlib/pull/265#discussion_r114299962) about its backward compatibility. Just wanted to share a simpler approach to fix some of the problems with auto refresh.

Lukasa:
Yeah, so I can understand that train of thought. However, this patch *also* subtly violates backwards compatibility, so my inclination is to go with the solution that advertises itself to users a bit more. Does that make sense?

Lukasa:
Didn't mean to close that, I'm apparently having a bad day for clicking on the correct buttons.

east825:
> my inclination is to go with the solution that advertises itself to users a bit more. Does that make sense?

Sure, all of a sudden I also realized that my approach with propagation of parameters like `timeout` and `proxies` to `refresh_token()` is additionally error-prone, because, in general, OAuth provider and protected resource accessed via REST API can be entirely different servers with different connection settings. 

Except for that ambiguity regarding the names of `auto_refresh_kwargs`, I agree that the solution from #265 is indeed more comprehensive and flexible. Feel free to close this PR.

Lukasa:
Thanks for the work!

PR #279, opened 2017-05-04
Avoid requests automatic auth

### LINKED ISSUE #278: OAuth2Session doesn't use token if netrc is present
The Heroku Toolbelt client uses a `.netrc` file to store its credentials locally. When using OAuth2Session, giving a properly-formed `token` to the constructor, the `auth` property on the session is not set, and requests goes and looks for a netrc file to add them in automatically when the request is made. When we _have_ set the token, this really needs to not happen.

The "ideal" fix would be to have the `OAuth2Session.auth` property set to `OAuth2`. I definitely don't know the complexities of how difficult this is, but this bug makes OAuth2Session unusable for me, and I'm having to drop down to using `OAuth2` auth directly instead for my use-case.

ryanhiebert:
Fixes #278 

I haven't added any tests, but I can later.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,7 +4,7 @@ History
 UNRELEASED
 ++++++++++
 
-nothing yet
+- Avoid automatic netrc authentication for OAuth2Session.
 
 v1.1.0 (9 January 2019)
 +++++++++++++++++++++++
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -74,6 +74,10 @@ class OAuth2Session(requests.Session):
         self.auto_refresh_kwargs = auto_refresh_kwargs or {}
         self.token_updater = token_updater
 
+        # Ensure that requests doesn't do any automatic auth. See #278.
+        # The default behavior can be re-enabled by setting auth to None.
+        self.auth = lambda r: r
+
         # Allow customizations for non compliant providers through various
         # hooks to adjust requests and responses.
         self.compliance_hook = {
--- a/tests/test_oauth2_session.py
+++ b/tests/test_oauth2_session.py
@@ -1,6 +1,9 @@
 from __future__ import unicode_literals
 import json
 import time
+import tempfile
+import shutil
+import os
 from base64 import b64encode
 from copy import deepcopy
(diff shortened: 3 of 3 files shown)


Lukasa:
So, this definitely looks plausible. I think we need some testing: in particular, I think we need some testing about how this interacts with the other places we process auth (see anywhere that has `auth=` in the codebase).

ryanhiebert:
I'll try to put in some more time this evening to figuring out the testing.

singingwolfboy:
Hey @ryanhiebert, what's the status of this pull request?

ryanhiebert:
Thank you for the nudge. This latest commit fixed the bugs, but I still haven't added any regression testing.

ryanhiebert:
I am working on the regression test, but looking at @Lukasa 's comment to me about checking on auth that we do, I'm not sure how to do any of that. My impression is that there's nothing really to test with that, because all of the current tests should handle those cases.

The only thing I'm seeing being worth adding AFA tests is the regression test to make sure that the `.netrc` isn't used, even though thats the default for requests.

ryanhiebert:
Digging into the session code of requests a bit more, it looks like there's a `trust_env` setting that defaults to `True`.

http://docs.python-requests.org/en/master/api/#requests.Session.trust_env

Would this be a better option than the noop auth? It'd certainly work for my use-case. The downside is that it looks like it also bypasses any proxies, so it's not strictly relegated to just avoiding using the `.netrc` files, though I'm not sure how much of an issue that caveat is.

ryanhiebert:
OK, I've added a regression test suite, that inherits from the main test suite but sets up a problematic `.netrc` file before it begins, and caused 2 of the tests to fail (so they are good regression tests).

I've kept the no-op auth as the approach, because I'm a bit hesitant to incur the side-effect of bypassing proxies, and because the no-op lambda is already short enough to make me happy.

singingwolfboy:
Looks great to me! @Lukasa, what do you think?

singingwolfboy:
@ryanhiebert if you rebase to resolve the conflicts, and the tests still pass, then I'll merge this.

ryanhiebert:
Thanks, @singingwolfboy . I've done that.

PR #283, opened 2017-07-10
Added "scope=None" argument to fetch_token

SNWBRS:
In my back-end application it's necessary to pass body as grant_type=client_credentials&scope=bla+blabla.
That is why it can be useful to pass optional argument "scope" to fetch_token and then to prepare_request_body.
Maybe someone will find it useful too.

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -152,7 +152,7 @@ class OAuth2Session(requests.Session):
                 state=state,
                 **kwargs), state
 
-    def fetch_token(self, token_url, code=None, authorization_response=None,
+    def fetch_token(self, token_url, scope=None, code=None, authorization_response=None,
             body='', auth=None, username=None, password=None, method='POST',
             timeout=None, headers=None, verify=True, proxies=None, **kwargs):
         """Generic method for fetching an access token from the token endpoint.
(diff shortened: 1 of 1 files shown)


Lukasa:
Is there a reason that the `scope` keyword argument to the `OAuth2Session` doesn't work for you here?

SNWBRS:
Yes, it does not change grant_type in a proper way:
grant_type=client_credentials
auth = HTTPBasicAuth(client_id, client_secret)
client = BackendApplicationClient(client_id=client_id)
oauth = OAuth2Session(client=client,scope=scope)
token = oauth.fetch_token(token_url=token_url, auth=auth)

DEBUG:requests_oauthlib.oauth2_session:Prepared fetch token request body grant_type=client_credentials

when i need grant_type=client_credentials&scope=bla+blabla

Lukasa:
@SNWBRS Sure, but can we not just change the patch to use `self.scope` rather than take a new `scope` argument? Is there any reason to assume they'll be different?

SNWBRS:
Off course `self.scope` will do for prepare_request_body

singingwolfboy:
Hey @SNWBRS, what's the status of this pull request? Any chance we can get it merged? Looks like you still need to add a changelog entry and an automated test.

PR #284, opened 2017-08-06
use prompt instead of approval_prompt in docs

aliva:
it seems google has change some of their APIs, ref: https://github.com/google/oauth2client/issues/453


Lukasa:
Thanks for this! :sparkles:

Issue #292, opened 2017-10-17
Client Certificates


amackera:
Hey there,

Right now it doesn't look like `requests-oauthlib` passes the `cert` argument on to `requests` during a `request`, or `fetch_token`. 

Maybe I'm missing how to do it? I'm happy to make a PR for this.

sserrata:
Hi @amackera, were you ever able to resolve this issue?

I've tried including a client certificate a number of ways and keep seeing the following error message from my IdP:

`(invalid_client) <client-id> is configured to use client certificates but there were no client certificates sent in the request.`

Oauth2Session doesn't accept the `cert` argument. Attempting to pass it results in the following exception:

`TypeError: __init__() got an unexpected keyword argument 'cert'`

The fetch_token method does accept the `cert` argument and the prepared fetch token body and request body appear to include it. Despite this, I'm still getting a 401 with the message mentioned above.

Note: I also tried incorporating your PR and still get the same results.

PR #293, opened 2017-10-17
Adds client certificate parameter to OAuth2Session

amackera:
This allows users to specify a client certificate to pass to `requests`. This is useful for some OAuth providers that require client certificate authentication.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,71 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+  schedule:
+    - cron: '33 8 * * 3'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'python' ]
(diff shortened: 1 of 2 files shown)


amackera:
Related issue: https://github.com/requests/requests-oauthlib/issues/292

Lukasa:
Requests should do this automatically: did you actually try using the `cert` argument in this way?

sserrata:
Has anyone validated this behavior? I've tried including a client certificate a number of ways and keep seeing the following error message from my IdP:

`(invalid_client) <client-id> is configured to use client certificates but there were no client certificates sent in the request.`

Oauth2Session doesn't accept the `cert` argument. Attempting to pass it results in the following exception: 

`TypeError: __init__() got an unexpected keyword argument 'cert'`

The fetch_token method does accept the `cert` argument and the prepared fetch token body and request body appear to include it. Despite this, I'm still getting a 401 with the message mentioned above.

amackera:
Hi @Lukasa yes, we tried using the `cert` argument, but it did not work. It's been a while since I looked at this issue though so my memory could be faulty! 

Feel free to close this PR if y'all know for a fact that `requests-oauthlib` supports client certificates out of the box.

ltamaster:
I had the same problem, I tried to use the `cert`argument without luck.
Any advice on how to pass the certificate?

Thanks
Luis

sserrata:
I've confirmed that the `cert` parameter works as either a `kwarg`, e.g. `fetch_token(cert='cert.pem')` or as an OAuth2Session() attribute, e.g. `oauth2session.cert = 'cert.pem'`.

ltamaster:
I tried `oauth2session.cert =tmp-cert-path`  and oauth2session.refresh_token(****, cert=tmp-cert-path) without luck, any ideas? 
the certificate works ok on other calls (i am working with kubernetes sdk)

calcium:
I just cloned the current trunk (master) and looking at the code for fetch_token
oauth2_session.py:219
       if method.upper() == 'POST':
            r = self.post(token_url, data=dict(urldecode(body)),
                timeout=timeout, headers=headers, auth=auth,
                verify=verify, proxies=proxies)

I dont see how this would work as is without the extra cert args. 

Yes, I need to do this too :-)

Anmol-Porwal18:
Hi, Is using oauth2session.cert the expected way to pass certificates ?

I need this functionality too (to be able to pass certificate in fetch_token - which will take care of passing it further). Are we expecting this to be merged soon ?

amackera:
Closing this (years old) PR due to lack of activity.

Issue #107, opened 2014-02-11
Trouble to make the Oauth2 Web App example working


ihommani:
Hello, 

I've got the following error when I try the snippet of [code](https://requests-oauthlib.readthedocs.org/en/latest/examples/real_world_example.html#web-app-example-of-oauth-2-web-application-flow)

**oauthlib.oauth2.rfc6749.errors.InsecureTransportError**
I got the redirection but returning to the callback throws this exception.

The exception is triggered by  the line :
token = github.fetch_token(token_url, client_secret=client_secret,                            authorization_response=request.url)

Since I'm new to requests-oauth, I don't know if the blunder is coming from me or not. But I almost copy/paste the code and replaced all that have to be replaced.

Can anyone confirm the example is working ?

hellupline:
there is a regression of this issue
oauthlib change their debug ENV var from DEBUG to OAUTHLIB_INSECURE_TRANSPORT for this exception

I will create a PR to update the example

PR #305, opened 2018-01-12
update oauthlib debug envvar with new variable

hellupline:
update examples with new envvar to suppress exception from oauthlib  
#107


singingwolfboy:
Thanks!

PR #310, opened 2018-04-10
OAuth1Session token property

singingwolfboy:
This pull request adds a `token` property to the `OAuth1Session` class, to match the `token` property on the `OAuth2Session` class. It also changes the `parse_authorization_response` and `_fetch_token` methods to call the setter for this property.

You may notice that these two methods now call `self._populate_attributes()` immediately followed by the `token` setter, which _also_ calls `self._populate_attributes()`. Why the duplication? Because subclasses may override this `token` property to behave differently, while still expecting the base class to populate its internal attributes. This how [Flask-Dance](https://github.com/singingwolfboy/flask-dance) works -- calling the `token` setter activates the token storage backend.

If you examine the `OAuth2Session` class in this project, you will see that this duplication appears there, as well. This pull request is about making `OAuth1Session` logically consistent with `OAuth2Session`.

See also this issue on the Flask-Dance repository, which was the impetus for creating this pull request: https://github.com/singingwolfboy/flask-dance/issues/120

### DIFF
--- a/requests_oauthlib/oauth1_session.py
+++ b/requests_oauthlib/oauth1_session.py
@@ -169,6 +169,26 @@ class OAuth1Session(requests.Session):
         self.auth = self._client
 
     @property
+    def token(self):
+        oauth_token = self._client.client.resource_owner_key
+        oauth_token_secret = self._client.client.resource_owner_secret
+        oauth_verifier = self._client.client.verifier
+
+        token_dict = {}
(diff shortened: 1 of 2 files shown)


sigmavirus24:
Closing & re-opening to trigger a rebuild

sigmavirus24:
:tada:

PR #312, opened 2018-05-16
Update HISTORY.rst

singingwolfboy:
This is in preparation for a new release -- I suggest that we make this 1.0, since this project has clearly been stable for awhile.


sigmavirus24:
:+1: on 1.0 since we're dropping support for versions of Python

PR #314, opened 2018-05-30
Compatibility to oauthlib v2.1.0

SunFulong:


requirements.txt
```diff
@@ -1,2 +1,2 @@
 requests>=2.0.0
-oauthlib[signedtoken]>=0.6.2
+oauthlib[signedtoken]>=2.1.0
```
singingwolfboy: According to [the backwards-compatibility pull request for oauthlib](https://github.com/oauthlib/oauthlib/pull/546), we only need to pin oauthlib to 2.0.8 or higher. Can you make this change?

Also, please modify the `install_requires` and `extras_require` section of the `setup.py` file to match this file.

singingwolfboy: Oh, I just checked the [CHANGELOG](https://github.com/oauthlib/oauthlib/blob/master/CHANGELOG.rst), and it looks like there is no 2.0.8 release -- 2.1.0 is the correct version. My mistake!

However, you still need to modify the `setup.py` file to match this `requirements.txt` file.

SunFulong: Modified setup.py, now it's ready to merge.

PR #325, opened 2018-08-12
fix Oauth2Session post json bug

tetrau:
OAuth2Session inherits requests.Session.post method which has two key word arguments - json and data -. But when calling OAuth2Session.post with a json payload, an expired token, and auto refresh enabled, the json argument will be passed to OAuth2Session.refresh_token and cause token refreshing failed. 

```python
>>> # something like this will raise an error
>>> session.post('https://...', json=data_obj)
...
oauthlib.oauth2.rfc6749.errors.InvalidClientIdError: (invalid_request) AADSTS90013: Invalid input received from the user.
...
>>> # but using the data argument works fine
>>> session.post('https://...', data=json.dumps(data_obj))
```
This commit add the json argument to theOAuth2Session.request, prevent json argument being captured into **kwargs and pass to the OAuth2Session.refresh_token.

The commit has passed the unittest and my own test.

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 import logging
+import json as json_lib
 
 from oauthlib.common import generate_token, urldecode
 from oauthlib.oauth2 import WebApplicationClient, InsecureTransportError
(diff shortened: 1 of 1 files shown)


tetrau:
It seems like that the travis-ci tests passed on python2.7 python3.4 python3.5 but failed on python 3.6 with the following output
```OCI runtime exec failed: exec failed: container_linux.go:348: starting container process caused "process_linux.go:86: executing setns process caused \"exit status 21\"": unknown```
Any advice ?

PR #326, opened 2018-08-14
Remove client_id and client_secret from body (generated via prepare_r…

DmitryPaschenko:
Remove client_id and client_secret from body (generated via prepare_request_body) if HTTPBasicAuth used.

Usage:

token = oauth.fetch_token(
    token_url='token_url_here',
    code='code_here',
    client_id='client_id_here',
    client_secret='client_secret_here'
)

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -192,16 +192,12 @@ class OAuth2Session(requests.Session):
                 raise ValueError('Please supply either code or '
                                  'authorization_response parameters.')
 
-
-        body = self._client.prepare_request_body(code=code, body=body,
-                redirect_uri=self.redirect_uri, username=username,
-                password=password, **kwargs)
-
-        client_id = kwargs.get('client_id', '')
         if auth is None:
+            client_id = kwargs.pop('client_id', '')
+
             if client_id:
                 log.debug('Encoding client_id "%s" with client_secret as Basic auth credentials.', client_id)
-                client_secret = kwargs.get('client_secret', '')
+                client_secret = kwargs.pop('client_secret', '')
                 client_secret = client_secret if client_secret is not None else ''
                 auth = requests.auth.HTTPBasicAuth(client_id, client_secret)
             elif username:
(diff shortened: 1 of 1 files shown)


JonathanHuot:
Hi @DmitryPaschenko, the problem in this is that it takes out `client_id` from kwargs but it can still be present in `self._client` property. So, if you use `OAuth2Session(client_id=xx)` it will still be taken in account.

@jvanasco is working on a wider patch in oauthlib https://github.com/oauthlib/oauthlib/pull/593 and requests-oauthlib (PR TBC).

lepture:
@JonathanHuot the concept in current implementation is not good to understand. Here are my thoughts:

There are two types of requests:

1. using client information to get token
2. using token to request other things

For "using token to request other things", currently, `requests-oauthlib` is using `auth` parameter which is great.

Why not use the `auth` feature in `fetch_token` too? Currently, it only happens in some cases like here https://github.com/requests/requests-oauthlib/blob/master/requests_oauthlib/oauth2_session.py#L291

---

From the later RFCs, we can understand fetching access token in this way:

> we use client authentication information to exchange access token.

There are several methods for client authentication:

1. none  (just client_id)
2. client_secret_basic (client_id and client_secret in header, using "basic" auth)
3. client_secret_post (client_id and client_secret in payload "body")

For fetching token, this is usually called `token_endpoint_auth_method`.

You can get some inspiration from [Authlib](https://github.com/lepture/authlib). Checkout

1. https://github.com/lepture/authlib/blob/master/authlib/oauth2/client.py
2. https://github.com/lepture/authlib/blob/master/authlib/oauth2/auth.py
3. https://github.com/lepture/authlib/blob/master/authlib/client/oauth2_session.py

When fetching token, we just pass an `auth` parameter, this `auth` will decide where to add the client_id or client_secret. For instance:

```py
auth = ClientAuth(client_id, client_secret, 'client_secret_basic')
``

This `auth` will add client_id, client_secret in header.

```py
auth = ClientAuth(client_id, client_secret, 'client_secret_post')
```

This `auth` will add client_id, client_secret in body.

jvanasco:
> Why not use the `auth` feature in `fetch_token` too?

unless I am missing something, [`fetch_token` already accepts an auth 
](https://github.com/requests/requests-oauthlib/blob/master/requests_oauthlib/oauth2_session.py#L181)

[It is defined as](https://github.com/requests/requests-oauthlib/blob/master/requests_oauthlib/oauth2_session.py#L208)

        :param auth: An auth tuple or method as accepted by `requests`.

the block you cited only creates a HTTP Basic Auth header if the auth parameter was not provided and the requires/is-intended-to-have the header.

lepture:
@jvanasco I mean, for `client_secret_post` and `none` we can also use `auth`.

jvanasco:
I am sorry for wildly misinterpreting your original message. I didn't have enough coffee this morning - but I understand it now. 

you should probably create a new issue ticket with your above content. I think this PR ticket is a candidate to be closed, because it should have been made obsolete by PR #593 last year.

PR #331, opened 2018-09-17
support for client_id and oauthlib PR #593 (Issue #585)

jvanasco:
This PR updates `requests_oauthlib` support to https://github.com/oauthlib/oauthlib/issues/585 and https://github.com/oauthlib/oauthlib/pull/593

Of importance:

* The logic in `fetch_token` was largely redone. The existing approach tried to support an array of different ways to provide client and user details which led to easily broken implementations. The new approach favors full RFC compliance by default, but still maintains ways to build broken clients and talk to broken servers.

* New tests were added to ensure the above functionality works.

* Some existing tests were changed. Notably calls to `fetch_token` in `OAuth2SessionTest` allowed the LegacyApplicationClient to invoke the endpoint without the required user credentials.  The tests now ensure an exception is properly raised, and then retest with necessary arguments added in.

This PR is waterfalled against a #593 and should not be reviewed until that is accepted.

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -190,7 +190,9 @@ class OAuth2Session(requests.Session):
                                   include (False).
         :param client_secret: The `client_secret` paired to the `client_id`.
                               This is generally required unless provided in the
-                              `auth` tuple.
+                              `auth` tuple. If the value is `None`, it will be
+                              omitted from the request, however if the value is
+                              an empty string, an empty string will be sent.
         :param kwargs: Extra parameters to include in the token request.
         :return: A token dict
         """
@@ -261,7 +263,7 @@ class OAuth2Session(requests.Session):
         if include_client_id:
             # this was pulled out of the params
             # it needs to be passed into prepare_request_body
-            if client_secret:
+            if client_secret is not None:
                 kwargs['client_secret'] = client_secret
 
         body = self._client.prepare_request_body(code=code, body=body,
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -4,6 +4,7 @@ import logging
 
 from oauthlib.common import generate_token, urldecode
 from oauthlib.oauth2 import WebApplicationClient, InsecureTransportError
+from oauthlib.oauth2 import LegacyApplicationClient
 from oauthlib.oauth2 import TokenExpiredError, is_secure_transport
 import requests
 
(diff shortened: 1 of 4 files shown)


skion:
> This PR is waterfalled against a #593 and should not be reviewed until that is accepted.

And in reverse, it would be great if someone here can review https://github.com/oauthlib/oauthlib/pull/593  before we merge that.

requests_oauthlib/oauth2_session.py
```diff
@@ -192,23 +209,66 @@ def fetch_token(self, token_url, code=None, authorization_response=None,
...
+
+        if isinstance(self._client, LegacyApplicationClient):
+            if username is None:
+                raise ValueError('`LegacyApplicationClient` requires both the '
+                                 '`username` and `password` parameters.')
+            if password is None:
+                raise ValueError('The required paramter `username` was supplied, '
+                                 'but `password` was not.')
+
+        # TODO:
```
JonathanHuot: As discussed in https://github.com/oauthlib/oauthlib/issues/585#issuecomment-421169638, `username` and `password` are used only for `LegacyApplicationClient`. However, we should not *restrict* or try to *remove* them from the params, if they are present. But we should not do any processing either, since they are not described in the RFC (for those other grants!).

jvanasco: Great. so just remove the `TODO` line that I put in for code review - correct?

JonathanHuot: Yes IMHO

jvanasco: done.

the way i was notified of this by code-highlighting, I thought there may have been another issue or mis-interpretation. I just wanted to ensure we are both talking about the TODO line.

JonathanHuot:
LGTM, need to have a review from @requests-oauthlib team now ;-)

jvanasco:
IIRC, this introduced a breaking change(s) against the live version of oauthlib.  The next version of oauthlib and this PR have tests to catch an improper usage pattern that trigger an exception, so tests will fail on Travis by design.  Last I checked, they all pass against the next oauthlib  (master branch), which has the mid set of changes merged already.

At least that's what I remember. It's been a while.  I'm on vacation through mid January so someone else working on oauthlib will have to address this.

singingwolfboy:
I just merged https://github.com/requests/requests-oauthlib/pull/341 earlier today, which should restrict compatibility to versions of oauthlib below 3.0. Once oauthlib 3.0 is out, then it will be easier to ensure that requests-oauthlib is compatible with it.

singingwolfboy:
Can you guys take a look at #357, and tell me what you think?

Issue #333, opened 2018-10-16
Should support Flask app.test_client()


aparamon:
Currently, `OAuth1Session`/`OAuth2Session` support `requests` exclusively, both for communicating with authorization server and for accessing protected resources. However it would be of much value to additionally support *`requests`-compatible* objects, for example Flask `app.test_client()`.
It will require ability to optionally specify different session objects for communicating with authorization server (`fetch_token`) and for accessing protected resources (`request`).

PR #335, opened 2018-10-18
Support separate session for communicating with resource servers

aparamon:
Partly fixes https://github.com/requests/requests-oauthlib/issues/333.


aparamon:
Calling for review!
The wrapping method in `oauth2_session()` might look a bit too hacky; would `__getattr__`-based wrapper be a better implementation? The latter has the benefit of not modifying the original session object, but each call gets +1 level of indirection.

aparamon:
Obsoleted by https://github.com/lepture/authlib/pull/101 and further refactorings in [Authlib](https://github.com/lepture/authlib).

PR #350, opened 2018-12-26
Drop redundant 'requests' in extras_require

jdufresne:
Already included in `install_requires`, which are always included in the requirements. There is no need to list it a second time.


jdufresne:
Rebased. :slightly_smiling_face:

PR #357, opened 2019-01-11
Switch to OAuthlib 3.0.0

### LINKED ISSUE #356: oauthlib-3.0.0 is released
requests-oauthlib is incompatible.  Mind raising the limit (if tests pass) and making a new release (1.1.1 at this point)?

oauthlib[signedtoken]>=2.1.0,<3.0.0

https://github.com/requests/requests-oauthlib/commit/1b9fe1a630eb1c91bf12fd70aa2e10ca74ffc0b6

singingwolfboy:
Fixes #356, builds on top of #331.

This pull request adds support for OAuthlib 3.0.0, and **drops support** for versions of oauthlib before 3.0.0. I made this pull request based on the code in #331, which I assume is reasonably correct, but I don't have a good way of testing it. I had to make some minor changes to get the tests to pass on Python 3.

I would appreciate some eyes on this. The bulk of the code belongs to @jvanasco.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,6 +5,15 @@ UNRELEASED
 ++++++++++
 
 - Avoid automatic netrc authentication for OAuth2Session.
+- Updated oauth2 tests to use 'sess' for an OAuth2Session instance instead of `auth`
+  because OAuth2Session objects and methods acceept an `auth` paramether which is
+  typically an instance of `requests.auth.HTTPBasicAuth`
+- `OAuth2Session.fetch_token` previously tried to guess how and where to provide
+  "client" and "user" credentials incorrectly. This was incompatible with some
+  OAuth servers and incompatible with breaking changes in oauthlib that seek to
+  correctly provide the `client_id`. The older implementation also did not raise
+  the correct exceptions when username and password are not present on Legacy
+  clients.
 
 v1.1.0 (9 January 2019)
 +++++++++++++++++++++++
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,7 +4,8 @@ History
 UNRELEASED
 ++++++++++
 
-- Avoid automatic netrc authentication for OAuth2Session.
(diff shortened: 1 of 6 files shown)


jvanasco:
thanks for doing this! sorry i missed that str/bytes issue in the test for `fake_refresh_with_auth`.

fyi, i wrote both the 'breaking changes' PR for oauthlib and the #331 code you built off of. there are some new test cases in there that should make you feel better.

the most general description of #331 is this: previous versions of both libraries did some stuff that was not spec compliant by default, and made it impossible to do other things that are spec compliant. with this patch, both projects should now default to spec compliance but can still do non-compliant things if needed.

singingwolfboy:
> fyi, i wrote both the 'breaking changes' PR for oauthlib and the #331 code you built off of. there are some new test cases in there that should make you feel better.

Oh, that's good to know! In that case, I'm very confident that the code you wrote actually works correctly. :) I'll give it a few more days to let others object if they want (although I don't expect anyone will), and then I'll merge this to master.

When it gets released, do you think it should be a major version bump (2.0.0), minor version bump (1.2.0) or micro version bump (1.1.1)? Assuming that there are no other changes in the release, of course.

jvanasco:
I think this should be fine as a minor bump. The only people who should expect problems are those who were relying on broken oauth2 implementations, because they:

* Are communicating with non-compliant servers which expect/require arguments on endpoints that do not require them (why oauthlib was modified).
* Have extensive tests which expect oauthlib/oauthlib-requests to generate URLs with certain arguments in them (why some tests in this library needed to be redone)

In both cases, it's because the oauth2 exchange was being implemented in a specially broken way, and oauthlib has exception messages that should properly educate the developer on what to do.

This "family" of PRs traces back to a regression in oauthlib where `client_id` was lost (https://github.com/oauthlib/oauthlib/issues/585) when expected. A chunk of code was rewritten to pass it around more correctly, and not re-introduce the error which the regression had solved. A few other spec-compliant checks and auto-population of arguments that interacted with `client_id` were also introduced to ensure valid requests.

Issue #363, opened 2019-01-27
`include_client_id`  auto detection


AdamGold:
I think that the "auto-detection" that is being done on `include_client_id` can be better. If no auth header is being passed, and include_client_id is `None`, I think we should pass the client ID and secret in the URL. What do you think?

PR #364, opened 2019-01-27
try detecting whether to add client id and secret to body

AdamGold:
I figured that if no auth header given AND `include_client_id` is not given, we probably do want to include client id and secret in the URL. #363

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -253,15 +253,17 @@ class OAuth2Session(requests.Session):
         else:
             # since we don't have an auth header, we MAY need to create one
             # it is possible that we want to send the `client_id` in the body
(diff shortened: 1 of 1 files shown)


singingwolfboy:
This pull request is failing the automated tests, so it can't be merged until that is resolved.

Also, can you explain why this change is a good idea? I don't actually know that much about this code, so I'm hesitant to make changes when I don't understand the context.

AdamGold:
@singingwolfboy 
Well, currently, if `auth` is None and `include_client_id` is None, `auth` is being created. I believe many people who upgraded to `1.2.0` had no idea about this `include_client_id` change, therefore didn't add it to their code. So I think the logic should be - if `auth` is None and `include_client_id` is None --> make `include_client_id` True. If `auth` is None and `include_client_id` is False --> create `auth`.

PR #361, opened 2019-01-27
add option to pass `params` in POST request

AdamGold:
- [x] add tests

PR #369, opened 2019-02-25
Add force_querystring argument to fetch_token() method

singingwolfboy:
Based on #361


singingwolfboy:
We can do that in a future pull request. I spent some time trying to figure out how to write tests for it, and I wasn't able to do it. The test harness needs a big overhaul.

PR #372, opened 2019-05-13
Removed redundancy

meyer1994:
`self.token` is a property. It has a setter that just calls `self._populate_attributes`.


singingwolfboy:
This duplication is actually intentional: see https://github.com/requests/requests-oauthlib/pull/310 for more details. 

> Subclasses may override this `token` property to behave differently, while still expecting the base class to populate its internal attributes. This how [Flask-Dance](https://github.com/singingwolfboy/flask-dance) works -- calling the token setter activates the token storage backend.

Would it help to add a comment to that code instead, to make it clearer that this duplication is intentional?

PR #374, opened 2019-06-09
Update to fix #373

### LINKED ISSUE #373: Error in if requests.__version__ < '2.0.0'
Hey there,
anyone ever got an Error like this:

`File "/discord_bot/redbot/cogs/Downloader/lib/twitter/api.py", line 31, in <module>
    from requests_oauthlib import OAuth1, OAuth2
  File "/discord_bot/redbot/cogs/Downloader/lib/requests_oauthlib/__init__.py", line 11, in <module>
    if requests.__version__ < '2.0.0':
TypeError: '<' not supported between instances of 'module' and 'str'`

I have no clue whats the problem. Maybe someone can help me ;)

entchen66:
https://github.com/requests/requests-oauthlib/issues

### DIFF
--- a/requests_oauthlib/__init__.py
+++ b/requests_oauthlib/__init__.py
@@ -9,7 +9,7 @@ __version__ = "1.2.0"
 
 import requests
 
-if requests.__version__ < "2.0.0":
+if str(requests.__version__) < "2.0.0":
     msg = (
         "You are using requests version %s, which is older than "
         "requests-oauthlib expects, please upgrade to 2.0.0 or later."


singingwolfboy:
It appears that in Requests 3, `__version__` is a module, not a variable: https://github.com/psf/requests/blob/master/requests/__version__.py Note that there is now a file called `requests/__version__.py`!

As a result, simply transforming this into a string is _not_ the correct way of doing this comparison. We're going to need to build out the automated test suite to handle a case where Requests 3 is installed. But considered that it's not even published yet, I don't think this work needs to happen yet -- it's possible that this problem will go away before Requests 3 is released.

PR #382, opened 2019-09-03
Fix LinkedIn OAuth

MerleLiuKun:
Hello, when I use the library to do LinkedIn OAuth, and I get some errors. I found that the OAuth has changed.

So I give this PR to solve it.


jtroussard:
@MerleLiuKun Thanks for all sharing your findings. These changes were incorporated with PR #397

Issue #377, opened 2019-08-16
access_token field added to request url when oauth2_access_token already present


jtroussard:
When trying to make a request for a LinkedIn resource I was recieving an unpermitted field error. 

> b'{"serviceErrorCode":100,"message":"Unpermitted fields present in PARAMETER: Data Processing Exception while processing fields [/access_token]","status":403}

Manual review of the URL revealed 2 access token related fields:

1. `oauth2_access_token`
2. `access_token`

Looking at the OAuth2-Requests source code in `requests-oauthlib/requests_oauthlib/oauth2_session.py`, the `access_token` field is added to the url right before making the final request via the `add_token()` method.

I imagine there is a mechanism in place to prevent the behavior but I could not find it? My solution was to copy a modified version of the `oauth2_session.py` module into my project with this dirty fix inside the `request()` method.

```python
old_version_url = url
url, headers, data = self._client.add_token(url, 
        http_method=method, body=data, headers=headers)
if "&access_token=" in url:
    url = old_version_url
```

Is there an official way to modify this behavior? Is this a bug? Please advise.

Thanks

tuseau:
It would be nice to have this fixed as it has completely broken the Linkedin integration.

jtroussard:
@tuseau You would think right? Turns out there's a Requests 3 that has been funded and under development. The main Requests websites mentions that this library is only in maintenance mode and that all attention is being given to Requests 3... Anyhoooo If your so inclined you can "fix" you LinkedIn integration by importing this modified copy of the [oauth2_session.py code](https://github.com/jtroussard/linkedin_assist/blob/master/linkedin_assist/quick_fixes/oauth2_session.py) into your module.

JonathanHuot:
@tuseau, @jtroussard, I think a PR can help the project !. However, as mentionned in https://github.com/requests/requests-oauthlib/issues/385, the overall delay could be a bit long. But that's better than not having it integrated at all.

Also about Requests 3 library, that would be great to improve requests-oauthlib to support it but that probably not an urgent needs for most users, because most of them are not using the Requests API anyway.

jtroussard:
This has been solved with PR #397

PR #388, opened 2019-09-24
Non compliant param name update/fix for LinkedIn integration.

jtroussard:
Work around to remove duplicate access token value from url in the request() method. 

Would be happy to collaborate on a more informed fix.

For more details please see issue:
[access_token field added to request url when oauth2_access_token already present #377](https://github.com/requests/requests-oauthlib/issues/377)

### DIFF
--- a/requests_oauthlib/compliance_fixes/linkedin.py
+++ b/requests_oauthlib/compliance_fixes/linkedin.py
@@ -11,8 +11,11 @@ def linkedin_compliance_fix(session):
         return r
 
     def _non_compliant_param_name(url, headers, data):
-        token = [("oauth2_access_token", session.access_token)]
-        url = add_params_to_uri(url, token)
+        if "&access_token=" in url:
+            url = url.replace("access_token", "oauth2_access_token")
+        else:
+            token = [("oauth2_access_token", session.access_token)]
+            url = add_params_to_uri(url, token)
         return url, headers, data
 
     session._client.default_token_placement = "query"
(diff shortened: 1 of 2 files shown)


JonathanHuot:
Is it related to linkedin only ? also, isn't better to not add a duplicate instead of removing once it has been added ?
Cheers

jtroussard:
@JonathanHuot I have not attempted to establish authentication with any resource provider besides LinkedIn. I was hoping someone more familiar with this library would be able to provide a little more insight as to how the module determines how/when/where to add the access token string to the request URL. Any ideas/advice as how to go about finding this out?

As for the present code, I was just thinking about that this morning. Your suggestion is more direct. I'll make the changes now. Thanks!

JonathanHuot:
Also, if it is strictely related to Linkedin integration, I think it deserves to be located in https://github.com/requests/requests-oauthlib/blob/master/requests_oauthlib/compliance_fixes/linkedin.py

Is it possible ?

Thanks

jtroussard:
@JonathanHuot The access token appending was in front of me this whole time. It looks like the "oauth_access_token" value is being added to the URL as part of the LinkedIn compliance hook routine. If that is the case I would purpose that the compliance hook `for` loop be moved after the "access_token" value is added to the URL and then have the `def _non_compliant_param_name()` method handle the field name vernacular. I'll make the necessary changes now.

jtroussard:
Any advice on why my commits are failing the file Travic CI check for Black? Is that a formatting thing? If it is I've run both modules through a Black formatting.

andr1an:
@jtroussard Yes, you can call `black requests_oauthlib/compliance_fixes/linkedin.py requests_oauthlib/oauth2_session.py` to fix this.

jtroussard:
This has been moved to #397

jtroussard:
This has been moved to #397

PR #389, opened 2019-11-29
Add Support for OAuth Mutual TLS (draft-ietf-oauth-mtls)

danielfett:
OAuth Mutual TLS is on track to become an IETF standard. It will be used in many high-risk scenarios, including Open Banking (see OpenID FAPI), e-health applications, and e-government applications. This patch adds support for client authentication via MTLS.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,7 +4,7 @@ History
 UNRELEASED
 ++++++++++
 
-nothing yet
+- Add initial support for OAuth Mutual TLS (draft-ietf-oauth-mtls)
 
 v1.3.0 (6 November 2019)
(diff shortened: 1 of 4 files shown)


singingwolfboy:
This looks like a great start! However, we need an automated test covering this change, and an entry in the `HISTORY.rst` file.

Let me know if you have any difficulty with writing this test. I think you'll want to mock the `requests` library, and verify that it's being called with the `cert` parameter.

danielfett:
Added tests and `HISTORY.rst` entry.

singingwolfboy:
Looks great! Thank you! 🎉

PR #394, opened 2020-01-09
Add support for auto-refreshing token without refresh token

### LINKED ISSUE #260: Possible to auto refresh client_credentials access token?
As a `client_credentials` based access token has no refresh token, the `refresh_token()` functionality raises an exception (ie. `InvalidRequestError: Missing refresh token parameter.`). Based on the spec, I believe this is *technically* fine - one should fetch a new access token in this case.

Is it just the spec that prevents `requests-oauthlib` from handling this scenario gracefully alongside the current token expiration code handling in the `request()` method on `OAuth2Session`?

If I subclass `OAuth2Session` like so:

```
class RefreshOAuth2Session(OAuth2Session):
    def request(self, *args, **kwargs):
        try:
            return super().request(*args, **kwargs)
        except TokenExpiredError:
            self.token = self.fetch_token(
                token_url=OAUTH2_TOKEN_URL,
                **self.auto_refresh_kwargs
            )
            self.token_updater(self.token)
            return super().request(*args, **kwargs)
```

I can use it like this:

```
class MyClass(object):
    _session = None
    _token = None

    def token_updater(self, token):
        self._token = token

    @property
    def session(self):
        if self._session is None:
            self._session = RefreshOAuth2Session(
                client=BackendApplicationClient(client_id=self.client_id),
                token=self.token,
                token_updater=self.token_updater,
                auto_refresh_kwargs={
                    'client_id': self.client_id,
                    'client_secret': self.client_secret,
                }
            )
        return self._session

    @property
    def token(self):
        if self._token is None:
            self._token = RefreshOAuth2Session(
                client=BackendApplicationClient(client_id=self.client_id)
            ).fetch_token(
                token_url=OAUTH2_TOKEN_URL,
                client_id=self.client_id,
                client_secret=self.client_secret
            )
        return self._token

    def __init__(self, *args, **kwargs):
        super(MyClass, self).__init__(*args, **kwargs)
        self.client_id = CLIENT_ID
        self.client_secret = CLIENT_SECRET

    def get_information(self):
        return self.session.get('https://localhost:10000/api/infomation/')


obj = MyClass()
obj.get_information()
```

As long as `auto_refresh_url` is not passed in, the logic in my subclass handles everything smoothly.  Could logic be added to `request()` to cover this scenario, rather than having to kludge this as I've done?

Related thread: https://groups.google.com/d/msg/django-oauth-toolkit/iSIiZQtn0mM/5SDckbyTCQAJ

denizdogan:
This is a basic idea on how to fix #260 since nothing has happened to the issue in three years. Let's get a discussion going, because this is a hindrance to me in almost every OAuth2 implementation I make using this library and it's clear that lots of other people have the same issue.

So the idea is to add a new constructor parameter which I currently call `auto_refresh_type`, it can be either `"refresh_token"` (which will do exactly what it does today) or `"access_token"` (which will just get a new access token as you normally do).

It's not a beautiful design by any means, but I intentionally tried to keep the diff as minimal as possible without any major refactorings, which would introduce breaking changes and make it more difficult to get this functionality out there.

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -41,6 +41,7 @@ class OAuth2Session(requests.Session):
         client=None,
         auto_refresh_url=None,
         auto_refresh_kwargs=None,
+        auto_refresh_type="refresh_token",
         scope=None,
         redirect_uri=None,
         token=None,
@@ -67,6 +68,8 @@ class OAuth2Session(requests.Session):
                            your access tokens.
         :auto_refresh_kwargs: Extra arguments to pass to the refresh token
                               endpoint.
+        :auto_refresh_type: Type of auto refresh method to use.  Must be either
+                            "refresh_token" (default) or "access_token".
         :token_updater: Method with one argument, token, to be used to update
                         your token database on automatic token refresh. If not
                         set a TokenUpdated warning will be raised when a token
@@ -83,6 +86,7 @@ class OAuth2Session(requests.Session):
         self._state = state
         self.auto_refresh_url = auto_refresh_url
         self.auto_refresh_kwargs = auto_refresh_kwargs or {}
+        self.auto_refresh_type = auto_refresh_type
         self.token_updater = token_updater
 
         # Ensure that requests doesn't do any automatic auth. See #278.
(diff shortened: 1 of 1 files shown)


JonathanHuot:
This looks amazing and probably a documentation about it and a couple of unit tests will be ideal...
I know it has been long time already @denizdogan. I can also try to do this part.

rohanliston:
This is exactly the functionality I'm looking for. Is there any chance of this being merged in the near future?

denizdogan:
@rohanliston Personally, I will not be working on adding docs or tests to this PR, I don't really use Python much anymore at all. If anyone really wants this to get merged, I would suggest maybe making a new PR with the same changes, plus some docs and a unit test.

rohanliston:
@denizdogan Thanks. I'm happy to submit a fresh PR with tests/docs, but before doing so I'd like to know if the maintainers are likely to review/merge it since there appear to be questions around the status of this project.

@JonathanHuot are you able to comment on the above?

JonathanHuot:
@rohanliston, I can review it when ready and integrate it when ready.

jtroussard:
@JonathanHuot @rohanliston i can take some time this Sunday to review.

rohanliston:
@jtroussard The project I was working on didn't go ahead and I'm now on something else. I don't think I'll have the bandwidth to come back to this, unfortunately.

jtroussard:
@rohanliston Thanks for the heads up.

jtroussard:
The torch has been passed to #526

Issue #404, opened 2020-04-07
`Session` subclasses should honor the `self.verify` flag by default


smarie:
I discovered this issue today while using Fiddler to debug my OAuth connection.

Each `requests.Session` object has a `self.verify` attribute stating if SSL certificates should be verified (for example to reject self-signed certificates).

Unfortunately, even if I set the attribute to `False` on an `OAuth2Session` object, this flag is not honoured because the default value is `True` (and not `None`) in `fetch_token` and `refresh_token`.
Note: it does not seem to happen with `OAuth1Session` objects.

I can provide a PR to fix this, it is extremely easy.

jtroussard:
By all means submit the PR and we'll get some folks to review it. Thanks!

PR #407, opened 2020-04-09
Have `OAuth2Session` correctly use the `self.verify` value if `verify` is not   overridden in `fetch_token` and `refresh_token`.

smarie:
Fixes issue #404


JonathanHuot (approved):
LGTM

smarie:
thanks!

PR #409, opened 2020-04-10
`OAuth2Session` constructor now uses its `client.scope` when a `client`   is provided and `scope` is not overridden.

### LINKED ISSUE #408: `OAuth2Session` constructor does not inherit or reuse the `client.scope` when a `client` is provided and `scope` is not overridden
```python
my_scopes = ['my_scope']
client = LegacyApplicationClient(..., scope=my_scopes)
oauth_session = OAuth2Session(client=client)

assert oauth_session.scope == my_scopes  # <-- AssertError
```

Unfortunately, in `OAuth2Session.refresh_token` the `self.scope` is used

I suggest to create a `scope` property which would return `self._scope` if overridden, or `self.client.scope` otherwise (if client is not None).

I'll propose a PR

smarie:
Fixes #408


smarie:
thanks !

PR #427, opened 2020-10-02
Don't pass 'files' through to token refresh

remko:
When using the `files` parameter in requests for multipart file uploads, and a token needs to be refreshed, the `files` are passed along with the refresh token request.

This pull request only passes `files` to the actual request (as is the case with `data`).


JonathanHuot:
LGTM

Issue #430, opened 2020-10-25
Amazon Cognito support?


space55:
Hello,

I just had a quick question. Are there any plans to provide support for Amazon Cognito?

They have an "oddity" with how they accept the client ID & secret on the TOKEN route: https://docs.aws.amazon.com/cognito/latest/developerguide/token-endpoint.html

> If the client was issued a secret, the client must pass its client_id and client_secret in the authorization header through Basic HTTP authorization. The secret is Basic Base64Encode(client_id:client_secret).

Last I checked, this wasn't standard, but it's also possible I'm completely misremembering the RFC.

Are there any plans to support this in the library? I can write a PR if needed.

Thanks!

jtroussard:
Please feel free to submit PR

Issue #379, opened 2019-08-21
Not possible to include auth in auto_refresh_kwargs


Freewolny:
Hello.
I feel that it should be possible to add `auth` (as supported by requests, in tuple form) to `auto_refresh_kwargs` in `OAuth2Session`. Right now both possible situations are bad:

- adding `auth` to `auto_refresh_kwargs` results in adding it to the body of the refresh request;
- adding `auth` to `OAuth2Session` directly would result in it being passed to all subsequent requests made with that session, which is not what I want.

I would like to add it to `OAuth2Session` in the way that would ensure that any automatic future calls to `refresh_token` will have `auth` passed as well, so it would end up being passed to requests and handled as HTTP Basic Auth.

OrangeDog:
`refresh_token` should be using the same auth (client id+secret) as `fetch_token`.
There isn't really any reason for it to be a separate function in the first place.

The refresh mechanism is unusable right now, as there's no way to provide the required auth.

OrangeDog:
You can hack around this by always passing `client_id` and `client_secret` to `Session.request()`.
They should only ever get used for automatic refresh.

GittCatt:
Is there anything wrong with my hack above?

OrangeDog:
This also also part of #264

JimHokanson:
What's the status on this? It seems like there may be solutions but I'm not smart enough to understand what is being proposed. I thought just setting the auth property to the basic authentication header would work but then of course subsequent API requests that need the access token as the authentication header fail. More details on my situation here:
https://stackoverflow.com/questions/65874797/include-authorization-in-a-oauth2session-for-requests-oauthlib

As is, this example in the docs (below) doesn't seem like it will actually work if the API server wants a basic authentication header for the refresh token.

https://requests-oauthlib.readthedocs.io/en/latest/oauth2_workflow.html#third-recommended-define-automatic-token-refresh-and-update
```
>>> from requests_oauthlib import OAuth2Session
>>> client = OAuth2Session(client_id, token=token, auto_refresh_url=refresh_url,
...     auto_refresh_kwargs=extra, token_updater=token_saver)
>>> r = client.get(protected_url)
```

timdawborn:
[...]

Issue #264, opened 2017-03-03
Problems with Token refreshing.


btimby:
First of all, thank you for this wonderful library. It works great and saved me tons of time.

I am having issues with one aspect: auto token refresh. I have identified 3 distinct problems, which I will describe below and for which I will suggest solutions. I will also reference this issue from multiple other related issues. I am happy to provide a PR once we can agree upon a suitable solution.

I derived my solution from the following, as my issue is a variant of the same issue:

https://github.com/requests/requests-oauthlib/issues/260

All of the problems start here:

https://github.com/requests/requests-oauthlib/blob/master/requests_oauthlib/oauth2_session.py#L338

```python
    auth = kwargs.pop('auth', None)
    if client_id and client_secret and (auth is None):
        log.debug('Encoding client_id "%s" with client_secret as Basic auth credentials.', client_id)
        auth = requests.auth.HTTPBasicAuth(client_id, client_secret)
    token = self.refresh_token(
        self.auto_refresh_url, auth=auth, **kwargs)
```

Issue 1
---------
If I pass an `auth` kwarg intended for the `request()` method, it is eradicated here. I am not sure if this is an issue in practice.

Issue 2
---------
Box, Google and Onedrive do not expect the `client_id` and `client_secret` to be provided via Basic authentication ala `Authorization` header. They instead expect these POSTed in the body. There is no way to customize this behavior. Also, I don't know what vendors expect what method here, I only know the 3 above as that is what I am working with.

Issue 3
---------
Any kwargs I pass which are intended for the `request()` method end up being passed to `refresh_token()` and are used when accessing the token endpoint. For example, if my original `request()` uploads a file, that file will be transferred to the token endpoint when attempting to refresh the token.

Issue 4
---------
`client_id` and `client_secret` are not normally passed to `request()` but are necessary for token refresh. Passing them is not a problem except for unintended consequences of them being passed along to `super().request()`

Proposed solution.
-----------------------
I think all of the above can be solved by making two changes.

1. Pass the `client_id` and `client_secret` to `__init__()` where they are stored as attributes (of `OAuth2Session` or underlying `WebApplicationClient`. In fact `client_id` is already handled this way, but `fetch_token()`, `refresh_token()` etc. do no utilize it and expect it passed again as a kwarg. This addresses issue 4.

[...]

PR #433, opened 2020-12-08
Adding refresh_token_request and access_token_request compliance hooks

violuke:
This is based upon #214, so thanks to @jimcortez for basically solving this for me :+1: :pray:

We needed this for a custom internal OAuth solution hence no included compliance fix example.

This solved the problem we're having and I believe will solve some or all of the points in issues #182, #379, #430 and #264 so should be useful for the wider community, despite not including a specific compliance fix example.

Here's our internal compliance fix:

```py
import base64
from oauthlib.common import add_params_to_uri

def internal_oauth_compliance_fix(session, client_secret: str):
    def _non_compliant_auth_header(url, headers, body):
        basic_auth_value = "{}:{}".format(session._client.client_id, client_secret)
        headers["Authorization"] = 'Basic {}'.format(base64.b64encode(basic_auth_value.encode("utf-8")).decode("utf-8"))
        token = [('token', session._client.access_token)]
        url = add_params_to_uri(url, token)
        return url, headers, body

    session.register_compliance_hook('refresh_token_request', _non_compliant_auth_header)
    return session
```

I hope this helps someone else, thanks again to @jimcortez for the original work and if this could be merged soon that would be much appreciated. :+1:


timdawborn:
An example of a large public application that needs Basic Auth for the refresh tokens is the Xero API: https://developer.xero.com/documentation/oauth2/auth-flow#refresh

violuke:
I've just also added support for an `access_token_request` hook. We were in the same situation as #244 (needing to send the data as json on a password grant type), but in our case the API was unable to accept via `application/x-www-form-urlencoded`. We also had to remove the `redirect_uri` key from the JSON or the request was rejected.

I know this is against the Oauth2 specs, but that's the whole point of compliant hooks :wink: Here's an example usage:

```python
def redacted_company_compliance_fix(session):
    def _request_in_json_not_form_urlencoded(url, headers, request_kwargs):
        headers["Content-Type"] = "application/json"

        if "redirect_uri" in request_kwargs["data"]:
            del request_kwargs["data"]["redirect_uri"]

        request_kwargs["json"] = request_kwargs["data"]

        del request_kwargs["data"]

        return url, headers, request_kwargs

    session.register_compliance_hook("access_token_request", _request_in_json_not_form_urlencoded)
    return session
```

violuke:
@jtroussard would be please be possible for this to be reviewed? I think it will help in lots of incompliant situations. Thank you 🙏

requests_oauthlib/oauth2_session.py
```diff
@@ -337,6 +338,10 @@ def fetch_token(
             request_kwargs["params"] = dict(urldecode(body))
         else:
             raise ValueError("The method kwarg must be POST or GET.")
+            
+        for hook in self.compliance_hook["access_token_request"]:
```
JonathanHuot, line 342: Is this access_token_request declared somewhere?

JonathanHuot (changes requested):
Would it be possible to add at least a unit test? This feature is a great addition, so thanks for the extra work.

requests_oauthlib/oauth2_session.py
violuke: Good point 🤦 I've corrected this.

I'm not sure the best way to go about a unit test for these as they don't change the response like the other compliance hooks. Any suggestions or advice would be appreciated. Thank you.

violuke:
@JonathanHuot I've added a test now for both of the new hooks. Thanks.

JonathanHuot:
Can you execute `tox -e black` and `tox -e py27` as they seem not passing in pipeline. Thanks in advance

violuke:
Both are sorted now, thanks.

JonathanHuot (approved):

JonathanHuot:
Thanks for this useful feature!

violuke:
@JonathanHuot would you please be able to make a release on PyPI for this so that we can switch our package manager over to the main project again? Thank you.

violuke:
Any chance of that PyPI release? 😉

JonathanHuot:
Hi @violuke, I will prepare a release this month. Thanks for your patience.

PR #435, opened 2021-01-16
Update documentation for Python 3

gschizas:
Given that Python 2 is out of support, the documenation examples should probably use Python 3 syntax (i.e. print as a function, input instead of raw_input) instead of Python 2 syntax.


JonathanHuot (approved):
LGTM Thanks for the update

jtroussard (approved):
lgtm will this trigger ci/cd - dont think so just trying

jtroussard:
Made a tiny whitespace chance to trigger the new coveralls/github actions pipeline.

PR #441, opened 2021-04-14
Raise HTTPError in case of HTTP 5XX responses.

ab:
It is very confusing to raise a MissingTokenError when the server has 
returned an HTTP server error.

Instead, raise requests.exceptions.HTTPError if the server has returned
an HTTP 5XX server error.

Prior PR conversation in PR #217 indicates that the maintainers do not
want to raise on 4XX errors due to certain providers using those
responses to send data. So we need a custom handler with a slight
variation on the built-in requests.models.Response.raise_for_status().

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -7,6 +7,7 @@ from oauthlib.oauth2 import WebApplicationClient, InsecureTransportError
 from oauthlib.oauth2 import LegacyApplicationClient
 from oauthlib.oauth2 import TokenExpiredError, is_secure_transport
 import requests
+from requests.exceptions import HTTPError
 
 log = logging.getLogger(__name__)
 
@@ -363,6 +364,8 @@ class OAuth2Session(requests.Session):
             log.debug("Invoking hook %s.", hook)
             r = hook(r)
 
+        self._raise_for_5xx(response=r)
+
         self._client.parse_request_body_response(r.text, scope=self.scope)
         self.token = self._client.token
         log.debug("Obtained token %s.", self.token)
@@ -448,6 +451,8 @@ class OAuth2Session(requests.Session):
         for hook in self.compliance_hook["refresh_token_response"]:
             log.debug("Invoking hook %s.", hook)
             r = hook(r)
+
+        self._raise_for_5xx(response=r)
 
         self.token = self._client.parse_request_body_response(r.text, scope=self.scope)
         if not "refresh_token" in self.token:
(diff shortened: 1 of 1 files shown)


ab:
@jtroussard @Lukasa @sigmavirus24 @singingwolfboy 

This could be considered a revised PR for #217.

I'm pretty sure tests for `pypy3.5-6.0` are failing on master. The error regarding Python cryptography and OpenSSL 1.0.2 doesn't seem related to my code changes at all.

Tests pass on all other Python versions.

ab:
> Please don't randomly ping people who don't work on the project

Sorry about that, you had weighed in on the linked PR.

Rjevski:
+1 for this - I'm seeing confusing exceptions when the upstream provider returns a generic error response (Nginx's bad gateway error page for example) so I would recommend merging this too.

The tests are failing for an unrelated error and would most likely fail on `master` too. Depending on how the project wants to fix this we could either upgrade to a newer OpenSSL version which the `cryptography` module now requires or set the `CRYPTOGRAPHY_ALLOW_OPENSSL_102` environment variable to skip the warning and try using the outdated OpenSSL anyway.

requests_oauthlib/oauth2_session.py
```diff
@@ -538,3 +543,40 @@ def register_compliance_hook(self, hook_type, hook):
...
+            # We attempt to decode utf-8 first because some servers
+            # choose to localize their reason strings. If the string
+            # isn't utf-8, we fall back to iso-8859-1 for all other
+            # encodings. (See psf/requests PR #3538)
+            try:
+                reason = response.reason.decode("utf-8")
+            except UnicodeDecodeError:
+                reason = response.reason.decode("iso-8859-1")
+        else:
+            reason = response.reason
```
Rjevski, line 569: You seem to only need the `reason` in the `elif 500 <= response.status_code < 600` block below, so maybe worth moving it in there? That way the behavior doesn't change for any other response codes (at the moment you try decode the reason even if ultimately you might not need it, thus introducing potential risk of breakage for no reason.

requests_oauthlib/oauth2_session.py
```diff
@@ -538,3 +543,40 @@ def register_compliance_hook(self, hook_type, hook):
...
+
+        elif 500 <= response.status_code < 600:
+            http_error_msg = "%s Server Error: %s for url: %s" % (
+                response.status_code,
+                reason,
+                response.url,
+            )
+
+        if http_error_msg:
+            raise HTTPError(http_error_msg, response=response)
```
Rjevski, line 582: You only define `http_error_msg` in the `elif 500 <= response.status_code < 600` block above, so maybe worth moving it up there (and then not explicitly checking for it nor assigning it to a variable and raising directly)?

requests_oauthlib/oauth2_session.py
ab: This code is from `requests.Response.raise_for_status()`, and I chose to make as few changes as possible since the only desired difference in behavior is to ignore 4XX errors.

deathiop:
Is this PR still being considered? I came across the same need and would love for this feature to move forward.

PR #444, opened 2021-04-28
make OAuth2 session picklable

bcavagnolo:


requests_oauthlib/oauth2_session.py
```diff
@@ -97,6 +93,11 @@ def __init__(
             "protected_request": set(),
         }
 
+    def auth(self, r):
```
satiowadahc, line 96: Could the same have been done with the lambda function?

satiowadahc:
This PR makes me hesitate on a security library when the top of the documentation says pickling is not secure.
Can you elaborate on the use case?

https://docs.python.org/3/library/pickle.html

requests_oauthlib/oauth2_session.py
Und3rf10w: That's how it is in the original code, but it makes the OAuth2 session not picklable...

darkrain42:
> This PR makes me hesitate on a security library when the top of the documentation says pickling is not secure. Can you elaborate on the use case?

The [multiprocessing library](https://docs.python.org/3/library/multiprocessing.html) (and consequently things such as concurrent.futures's [ProcessPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor)) requires that the objects be picklable.

darkrain42:
That said, the fix here is incomplete, at least with current versions of requests. The requests Session defines [a specific list of attributes to save/restore](https://github.com/psf/requests/blob/2a6f290bc09324406708a4d404a88a45d848ddf9/requests/sessions.py#L374-L387), which doesn't include the OAuth2Session's `_client` attribute, so the unpickled session is unusable when attempting to make a request.

For context, without the fix in the PR here, I see the following exception:

```
Traceback (most recent call last):
...
  File "/usr/lib/python3.11/multiprocessing/pool.py", line 873, in next
    raise value
  File "/usr/lib/python3.11/multiprocessing/pool.py", line 540, in _handle_tasks
    put(task)
  File "/usr/lib/python3.11/multiprocessing/connection.py", line 205, in send
    self._send_bytes(_ForkingPickler.dumps(obj))
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
AttributeError: Can't pickle local object 'OAuth2Session.__init__.<locals>.<lambda>' 
```

PR #462, opened 2021-11-11
Fix Sphinx error for oauth1 fetch_token documentation

JonathanHuot:
Fix https://requests-oauthlib.readthedocs.io/en/v1.3.0-docs/api.html#requests_oauthlib.OAuth1Session.fetch_request_token
docstring issue resulting in an absent "returns" field.

### DIFF
--- a/docs/examples/outlook.rst
+++ b/docs/examples/outlook.rst
@@ -1,5 +1,5 @@
 Outlook Calendar OAuth 2 Tutorial
-==========================
+=================================
 
 Create a new web application client in the `Microsoft Application Registration Portal`_
 When you have obtained a ``client_id``, ``client_secret`` and registered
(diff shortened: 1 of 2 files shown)


jtroussard (changes requested):
Sorry. Help me understand the fix? Is this just whitespace formatting?

JonathanHuot:
Exactly. Since the code is part of a list of params, it cannot have the same indent, else the list is considered as closed, and the parameters below (here it is `:returns:` is considered as part of text block).

DocString:

```
 :param realm: A list of realms to request access to.
 :param \*\*request_kwargs: Optional arguments passed to ''post''
 function in ''requests.Session''
 :returns: The response in dict format. 
```
Results in:

image

And if we indent the param description like below:

```
 :param realm: A list of realms to request access to.
 :param \*\*request_kwargs: Optional arguments passed to ''post''
     function in ''requests.Session''
 :returns: The response in dict format. 
```

It becomes correctly formatted:

image

jtroussard (approved):
lgtm

PR #469, opened 2021-11-23
WIP: Fix Travis Pipeline

JonathanHuot:
TODO list:
- fix travis
- fix coveralls
- autorelease when tagging

Issue #460, opened 2021-11-09
Update pipeline/Github actions


jtroussard:
Seems the coverage report and tests are not kicking off. Would like to update/fix this as part of the [1.3.1](https://github.com/requests/requests-oauthlib/milestone/3) milestone.

JonathanHuot:
I would consider this work as critical as it will help other PRs to be merged into master.

jtroussard:
@JonathanHuot Looks like the coveralls and branch protection features are running. I know we had a bit of a discussion about the way the coverall coverages were being calculated but can we confirm were we wanted to cut a release? Do we want to wait for a more substantial test of the branch protection features/coveralls coverage check? Do we want to include the documentation reformat/update into this milestone?

jtroussard:
@JonathanHuot Looks like 1.3.1 is wrapping up nicely. Are we ready to cut a release?

JonathanHuot:
@jtroussard , yes let's cut the release!

PR #470, opened 2021-11-24
Move from Travis to GitHub Actions 

JonathanHuot:
As discussed in https://github.com/requests/requests-oauthlib/issues/460 and in https://github.com/requests/requests-oauthlib/pull/469, it seems simpler to use GitHub actions.


JonathanHuot:
I tested AndreMiras/coveralls-python-action, but found out important issue here https://github.com/AndreMiras/coveralls-python-action/issues/13

Coverage is passing and multiple github checks are added to the report, which is good. However, the data is pushed to coveralls as a "new coverage", i.e. we don't know if the coverage has increased or decreased compared to the master branch (base).

See as below 
image


I will move back to coverallsapp/github-action@master which was requiring a lcov format and not a standard format that coverage produce.

JonathanHuot:
So with the AndreMiras/coveralls-python-action I get https://coveralls.io/builds/44520842.
Limitation: I though https://github.com/AndreMiras/coveralls-python-action/issues/13 was a problem, but it seems I have the same behavior with the other one. I'm not sure at this stage if how can it be fixed.

When using the official coverallsapp/github-action I get https://coveralls.io/builds/44524845.
Limitation: CI is more complex due to the format required lcov
Limitation: python2.7 disabled 
Limitation: a change in the way coverage is calculated for property/setter compared to base.

jtroussard:
> I tested AndreMiras/coveralls-python-action, but found out important issue here [AndreMiras/coveralls-python-action#13](https://github.com/AndreMiras/coveralls-python-action/issues/13)
> 
> Coverage is passing and multiple github checks are added to the report, which is good. However, the data is pushed to coveralls as a "new coverage", i.e. we don't know if the coverage has increased or decreased compared to the master branch (base).
> 
> See as below image
> 
> I will move back to coverallsapp/github-action@master which was requiring a lcov format and not a standard format that coverage produce.

Would the coverage comparison data not be shown if opening the "COVERAGE CHANGED" tab?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,62 @@
...
+      env:
+        TOXENV: ${{ matrix.tox-env }}
+      run: |
+        tox
+    - name: Coverage format into lcov
+      if: ${{ matrix.python-version != '2.7' }}
+      run: |
+        coverage-lcov --output_file_path lcov.info
+    - name: Coveralls Parallel
+      uses: coverallsapp/github-action@master
```
jtroussard, line 46: Not sure if this matters, but in my own repos I've had less problems when setting this value with an actual version number, I think 1.1.3 is latest stable:

```suggestion
      uses: coverallsapp/github-action@1.1.3
```

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,62 @@
...
+        github-token: ${{ secrets.github_token }}
+        flag-name: run-${{ matrix.python-version }}
+        path-to-lcov: lcov.info
+        parallel: true
+  finish:
+    needs: tests
+    runs-on: ubuntu-latest
+    steps:
+    - name: Coveralls Finished
+      uses: coverallsapp/github-action@master
```
jtroussard, line 58:
```suggestion
      coverallsapp/github-action@1.1.3
```

jtroussard (changes requested):
Could we also inject a small doc/setup/getting started readme, specifically with the `tox` tool. I messed around with it for a while and could only get py2.7 stuff to work. I'm sure I'm screwing something up with the setup on my end.

### UPDATE
So for potential documentation, I found that running `tox -e <python-version-youre-working-with>` simplifiled the local replication process. Also when running manually, be sure to install both requirements files (there are three, depends on which version of python you're running in venv) ... `pip install -r requirements.txt && pip install -r requirements-test.txt` AND `pip install coveralls coverage-lcov toml`

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,62 @@
...
+      with:
+        github-token: ${{ secrets.github_token }}
+        path-to-lcov: lcov.info
```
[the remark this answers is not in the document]
jtroussard: Ok nevermind mind this comment I misread one of the conditionals in the gh-actions workflow file and had a missing dependency when I was trying to rebuild this process by hand. Project root is the correct location.

```
    - name: Coverage format into lcov
      if: ${{ matrix.python-version != '2.7' }}
      run: |
        coverage-lcov --output_file_path lcov.info
```

jtroussard:
I think we can safetly add the following dependencies in `requirements-test.txt`:

- [ ] coverage-lcov
- [ ] toml

JonathanHuot:
> Could we also inject a small doc/setup/getting started readme, specifically with the `tox` tool. I messed around with it for a while and could only get py2.7 stuff to work. I'm sure I'm screwing something up with the setup on my end.
> ### UPDATE
> 
> So for potential documentation, I found that running `tox -e <python-version-youre-working-with>` simplifiled the local replication process. Also when running manually, be sure to install both requirements files (there are three, depends on which version of python you're running in venv) ... `pip install -r requirements.txt && pip install -r requirements-test.txt` AND `pip install coveralls coverage-lcov toml`

I have added a contributing guide, can you verify? Thanks!

JonathanHuot:
I'm parking the coverage question as of now, as we will need to see it in action with a couple of PRs to see if we can live with it or not. Anyway it should not be a blocker to start reviewing some PRs.

The PR is ready to be merged if review is OK!

jtroussard (approved):
lgtm

jtroussard:
looks good - let's confirm Github actions are working after this merge

Issue #443, opened 2021-04-21
python 3.8


toby221:
im on the newest version, python 3.8 but yours only supports up to 3.7, 

- [ ] could there be a fix?

hugovk:
Will be fixed by https://github.com/requests/requests-oauthlib/pull/442.

jtroussard:
#442 merged, closing this issue.

PR #442, opened 2021-04-20
Add Python 3.8 & 3.9 as supported versions

kaxil:
Since Py 3.8 is supported, we should update the "classifiers", as it is shown on PyPI and gives wrong impressions that Py 3.8 is not supported.

This PR also adds Py 3.9 and adds them to CI

PR #474, opened 2022-01-02
Add support for Python 3.10

hugovk:
Follow on from https://github.com/requests/requests-oauthlib/issues/443 / https://github.com/requests/requests-oauthlib/pull/442.

Python 3.10 was released on 2021-10-04:

* https://discuss.python.org/t/python-3-10-0-is-now-available/10955


JonathanHuot:
Thanks for update

Issue #473, opened 2022-01-01
There is a problem with python 3.8 and escape sequence:


jtroussard:
There is a problem with python 3.8 and escape sequence:
'invalid escape sequence \\*

```
________________________ ERROR collecting test session _________________________
/opt/virtualenvs/py38/lib/python3.8/site-packages/py/_path/common.py:383: in visit
    for x in Visitor(fil, rec, ignore, bf, sort).gen(self):
/opt/virtualenvs/py38/lib/python3.8/site-packages/py/_path/common.py:435: in gen
    for p in self.gen(subdir):
/opt/virtualenvs/py38/lib/python3.8/site-packages/py/_path/common.py:435: in gen
    for p in self.gen(subdir):
/opt/virtualenvs/py38/lib/python3.8/site-packages/py/_path/common.py:424: in gen
    dirs = self.optsort([p for p in entries
/opt/virtualenvs/py38/lib/python3.8/site-packages/py/_path/common.py:425: in <listcomp>
    if p.check(dir=1) and (rec is None or rec(p))])
/opt/virtualenvs/py38/lib/python3.8/site-packages/_pytest/main.py:667: in _recurse
    ihook = self.gethookproxy(dirpath)
/opt/virtualenvs/py38/lib/python3.8/site-packages/_pytest/main.py:482: in gethookproxy
    my_conftestmodules = pm._getconftestmodules(fspath)
/opt/virtualenvs/py38/lib/python3.8/site-packages/_pytest/config/__init__.py:431: in _getconftestmodules
    mod = self._importconftest(conftestpath.realpath())
/opt/virtualenvs/py38/lib/python3.8/site-packages/_pytest/config/__init__.py:470: in _importconftest
    raise ConftestImportFailure(conftestpath, sys.exc_info())
E   _pytest.config.ConftestImportFailure: (local('/conftest.py'), (<class 'SyntaxError'>, SyntaxError('invalid escape sequence \\*', ('/opt/virtualenvs/py38/lib/python3.8/site-packages/requests_oauthlib/oauth1_session.py', 258, 9, '        """Fetch a request token.\n')), <traceback object at 0x7fee844d6140>))
```

_Originally posted by @matejsp in https://github.com/requests/requests-oauthlib/issues/443#issuecomment-954664334_

hugovk:
Cannot reproduce locally:

```console
$ tox -e py38
GLOB sdist-make: /private/tmp/requests-oauthlib/setup.py
[...]

PR #478, opened 2022-01-02
Updates docs in oauth_session1 to match existing pattern

jtroussard:
## Notice
Accidentally closed last PR - trying to work out why coverage decreases by -0.0 percent when running coveralls with python version 3.7> 

## Original Description
Related to discussion on #473 #443 #442

Looks like all the uncertainty stems from a slight break for the existing documentation convention/pattern. Method documentation for, fetch_request_token in requests_oauthlib/oauth1_session.py added the double asterisk to the optional argument in the function description. The existing pattern is to add them for the signature and leave them off in the description. Before the doc string was flagged as a raw string, this was generating a warning for Invalid Escape Sequence.

Removing the raw string flag and the asterisks as it breaks the pattern already set.

### DIFF
--- a/requests_oauthlib/oauth1_session.py
+++ b/requests_oauthlib/oauth1_session.py
@@ -33,6 +33,7 @@ class TokenRequestDenied(ValueError):
         super(TokenRequestDenied, self).__init__(message)
         self.response = response
 
+    @property
     def status_code(self):
         """For backwards-compatibility purposes"""
         return self.response.status_code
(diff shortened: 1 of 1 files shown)


jtroussard:
Only difference I can find between 3.8> coveralls checks and the earlier versions of python is the calculation of relevant lines.[ Created issue in the coverallsapp repo hoping for some support](https://github.com/coverallsapp/github-action/issues/108). Anyone else see something I am missing?

JonathanHuot:
Merging as we know coverage is OK.

PR #480, opened 2022-01-21
Release 1.3.1

JonathanHuot:
After this MR is OK, we can tag the git commit, and it should (hopefully) automatically push the python package to pypi.
Have a look on changelog & new pipeline.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -1,0 +1,26 @@
+name: Publish Tags on PyPi when GitHub creates a release
+on:
+  push:
+    tags:
+      - 'v**'
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+      fail-fast: false
+      python-version: 3.9
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install upload dependencies
+      run: |
+        python -m pip install --upgrade pip setuptools wheel twine
+    - name: Build and publish
+      env:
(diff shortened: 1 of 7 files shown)


JonathanHuot:
Note that I avoided adding the automatic Github Release publishing, as I have too many Github Actions related to this... I was not able to choose, as none of them seems complete!

Basically, I was trying to replace the lovely TravisCI "releases" provider:

```
deploy:
  - provider: releases
    api_key:
      secure: "<..>"
    on:
      tags: true
      all_branches: true
      condition: $TOXENV = py36
      repo: requests/requests-oauthlib

```

JonathanHuot:
Pushing tag triggered the release in pypi as expected ! 
https://pypi.org/project/requests-oauthlib/1.3.1/

jtroussard:
I think we need to rethink our code coverage check, there is a bug I think on coverall side that causing these -0% coverage changes and creates these merge blocks. These are a few options:

- workaround this and increase request-oauthlib coverall threshold to something like 0.01%
- don't check all versions of python for coverage - should only really check for maintained and/or supported versions (this could still leave us with a version that is returning the -0% change)
- Only run coverage against the latest version of python

https://github.com/lemurheavy/coveralls-public/issues/1608#issuecomment-1005177210

@JonathanHuot

jtroussard:
@JonathanHuot - Circling back to this one here. Just needs a nudge to get it across the finish line. Who has access to the coveralls settings? Can we bump the coverage check threshold to 0.01% and run the pipeline against this release again?

JonathanHuot:
thanks to @nateprewitt, the coverage threshold has been changed to 1%, and now all checks are passing!

@jtroussard, the story about restoring pipeline is done finally. Now we can focus on bug & features PR reviews..

PR #485, opened 2022-02-21
Revert 397 and add Deprecation warning instead

jtroussard:
Semantic version mistake was made when releasing 1.3.1, the Linkedin compliance fix which was removed from the code base, api breaking change, did not trigger a main version release increment. e.g. 1.4 instead of 1.3.1.

This PR reverts the Linkedin compliance hook removal, and instead adds a deprecation warning. Can remove next major version release. Return Linkedin example file to original state as well as bringing back unit tests.

Updates HISTORY and AUTHOR file.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -25,3 +25,4 @@ Patches and Suggestions
 - Sylvain Marie <sylvain.marie@se.com>
 - Craig Anderson <craiga@craiga.id.au>
 - Hugo van Kemenade <https://github.com/hugovk>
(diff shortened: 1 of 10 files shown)


JonathanHuot (changes requested):
Can you rebase from 1.3.1 tag ?
That way we're not introducing other breaking changes already in master.

jtroussard:
@JonathanHuot we good now?

JonathanHuot:
Closed as no time to fix backward the version to maintain version semantic.

PR #487, opened 2022-03-17
fix: issue 411

hiporox:
It appears that if a `token_updater` is not specified, the client will not get updated. Instead of providing a default `token_updater`, we can instead move this client update call to happen after refresh regardless


AdrienKnell:
UP. Do we have any updates about this PR ? Can someone merge it ?

jtroussard:
Pushing a small change to kick off the PR build checks and tests

jtroussard:
Looks like all the checks have passed, is it possible to get/produce some evidence of integration or manual smoke/regression tests?

PR #522, opened 2023-12-04
Put OAuth2Session in 'oauth' consistently in examples

PaulWay:
Throughout the Worker example code, the `client` variable is assigned an `ApplicationClient` derived class object (e.g. BackendApplicationClient), and the `oauth` variable is assigned an `OAuth2Session` object.  However, in the 'Refreshing tokens' section the `client` is assigned an `OAuth2Session` object.  This change updates the latter section to use the variable convention of the former sections.


JonathanHuot:
Thank you for the suggestion and MR

PR #523, opened 2024-01-06
Removed older python versions (2.7,<=3.7)

JonathanHuot:
This should be the first step to 1.4.0 release, and a refresh of python versions. The codebase hasn't been cleaned, but it can come at a later point in time, or gradually depending the files changes.

The more important is to keep the latest tools (tox, cicd) up to date to allow MR to be included and release more often.


dosisod:
@JonathanHuot we should be able to remove the `requirements-test-27.txt` file as well right?

PR #525, opened 2024-01-13
Update Python 2 examples in docs to Python 3

dosisod:
I recently came across this library and saw that it still had Python 2 references in the docs. I saw #523 was open, so I went ahead and updated the documentation to use Python 3 examples.

Once #523 is merged this could be merged as well.

---

*(Out of scope, but leaving this here for now, could make this a separate issue)*

Now that we're dropping Python <=3.7 there are a lot of Python 3 related changes and quality of life improvements we can now make:
* Add type annotations
* Remove remaining Python 2 compatibility checks
* Improve testing infrastructure

I wouldn't mind working on these in my free time if we decide these would be useful!


JonathanHuot:
Will be happy to merge any code drop of py2

PR #532, opened 2024-03-06
Add option to disable logging of request/response headers at DEBUG level

erlendvollset:
Currently access tokens are exposed in logs when running with DEBUG log level. In many cases it's desirable to run a process with log level DEBUG without exposing such secrets. If there is some other way you think this should be configured, please advise.

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -47,6 +47,7 @@ class OAuth2Session(requests.Session):
         state=None,
         token_updater=None,
         pkce=None,
+        unsafely_log_headers_on_debug_level=True,
         **kwargs
     ):
         """Construct a new OAuth 2 client session.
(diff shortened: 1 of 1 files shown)


jtroussard:
I can take a closer look this weekend at this PR. At first blush though it will need an accompanying test or updates to existing tests.

jtroussard (changes requested):
The overall concept of controlling the level of logging in DEBUG mode is interesting and I can think of a few uses cases that would benefit from this sort of control. 

Considering implementation, it might be more fitting to integrate these controls directly with the logger to maintain separation of concerns. 

For example, introducing a filter to the logger that adjusts based on DEBUG level could be a cleaner less hardcoded approach. Additionally, offering options like NONE, MASKED, FULL for log visibility could add flexibility and keep a certain level of logging without causing security concerns. 

While the current proposal has merit, incorporating these adjustments could enhance functionality and align with coding best practices, all with a low LOE.

jtroussard:
First, I want to genuinely thank you @erlendvollset for your contribution and for spotlighting this security concern within the project. Your initiative is greatly appreciated. At the risk of stepping on toes, and after much consideration, I would like to propose and directly offer a different approach, which may address the issue more comprehensively. The nature of the changes I'm contemplating significantly diverges from the current proposal, extending beyond what might be practical to suggest via this PR alone or a simple PR code suggestion. To properly ilustrate this alt implementation I will open another PR inspired by this one (the commits will name you as a co-author). This is done with the utmost respect for your initial effort and is aimed solely at following a certain coding principles the community is trying to enforce. I hope this is taken in the collaborative spirit intended, and I'm more than open to discussing this further, potentially on even a different implementation all together.

PR #542, opened 2024-03-26
541: updates pypi metadata in setup file

jtroussard:
Updates setup.py classifiers to match recent updates which drop support for older versions of python.

Left the broader version 3 without a minor version declared for visibility.

Release with 2.x.x


singingwolfboy (approved):

JonathanHuot:
LGTM, which milestone/version should we add this release? :)

EpicWink:
My non-comprehensive testing does suggest that `requests-oauthlib` v2.0 works with Python 3.6, even if it's not supported and tested in CI, so it may not strictly be necessary to set `python_requires=">=3.7"`. I personally don't mind either way.

jtroussard:
> LGTM, which milestone/version should we set ? :)

This feels like a patch.

jtroussard:
> My non-comprehensive testing does suggest that `requests-oauthlib` v2.0 works with Python 3.6, even if it's not supported and tested in CI, so it may not strictly be necessary to set `python_requires=">=3.7"`. I personally don't mind either way.

Perhaps there is a disconnect between what "works" and what is actively being supported??? 

Is there any appetite to pursue this 3.6 verification beyond @EpicWink 's efforts?

EpicWink:
> Perhaps there is a disconnect between what "works" and what is actively being supported???

I think that's exactly the case. If you want to avoid complaint GitHub issues in the future, I suggest keep the `python_requires=">=3.7"` change currently in the PR, but if you want to let users stray into the territory of the unsupported (and are happy to close occasional issues on the topic), then leave it.

I personally only update the python-requires metadata when my packages start to use syntax/functionality only available in later versions; pip should handle dependency resolution on transitive dependencies' python version requirements.

Issue #544, opened 2024-04-19
Issue refreshing Wix OAuth tokens - cannot send payload as JSON using `refresh_token_request` compliance hook


skray:
**The crux of my question below is:** Am I correct that there is currently no way to use the `refresh_token_request` to send a JSON payload to the token request endpoint?

To explain the issue I am having, Wix's OAuth flow requires access token requests and refreshes to be made using (non-standard) JSON request bodies, instead of the standard `x-www-urlformencoded` body. Documentation for reference:

* Access token request: https://dev.wix.com/docs/rest/app-management/oauth-2/request-an-access-token
* Access token refresh: https://dev.wix.com/docs/rest/app-management/oauth-2/refresh-an-access-token

I am able to get the initial access token using the `access_token_request` compliance hook, using a hook very similar to [the one described in the PR](https://github.com/requests/requests-oauthlib/pull/433#issuecomment-1013384018) that introduced the it:

```python
def redacted_company_compliance_fix(session):
    def _request_in_json_not_form_urlencoded(url, headers, request_kwargs):
        headers["Content-Type"] = "application/json"

        if "redirect_uri" in request_kwargs["data"]:
            del request_kwargs["data"]["redirect_uri"]

        request_kwargs["json"] = request_kwargs["data"]

        del request_kwargs["data"]

        return url, headers, request_kwargs

    session.register_compliance_hook("access_token_request", _request_in_json_not_form_urlencoded)
    return session
```

However, I am not able to figure out a way to use the `refresh_token_request` compliance hook in a similar way to send a JSON payload. The [refresh logic](https://github.com/requests/requests-oauthlib/blob/45eca1a69dc2d1d9ec2d6769909a5a71e3548780/requests_oauthlib/oauth2_session.py#L478) is much more tightly coupled to sending the request payload as `x-www-form-urlencoded`, sending only explicit kwargs into the `self.post` method, including an expectation that the `body` returned from the `refresh_token_request` hook is a urlencoded string. Here is the code in question:

```python
for hook in self.compliance_hook["refresh_token_request"]:
    log.debug("Invoking refresh_token_request hook %s.", hook)
    token_url, headers, body = hook(token_url, headers, body)

r = self.post(
    token_url,
    data=dict(urldecode(body)),
    auth=auth,
    timeout=timeout,
    headers=headers,
    verify=verify,
    withhold_token=True,
    proxies=proxies,
)
```

So:
[...]

PR #545, opened 2024-04-19
Support non-x-www-form-urlencoded bodies returned from refresh_token_request compliance hook

skray:
Addresses https://github.com/requests/requests-oauthlib/issues/544

Adding similar capabilities to the `refresh_token_request` compliance hook as is possible in the `access_token_request` hook, to allow for non standard, non `x-www-form-urlencoded` request bodies in token refresh requests.

Additionally, adding in a Wix compliance fix showing how to use the `access_token_request` and `refresh_token_request` compliance hooks to successfully request and refresh tokens with Wix. Docs here: https://dev.wix.com/docs/rest/app-management/oauth-2/request-an-access-token 

I added tests for the compliance fix to show the `refresh_token_request` updates work (as well as testing them manually against the Wix API from my own project). I'm happy to add tests in `test_oauth2_session.py` as well, I just wasn't sure exactly what to add, as the code I added should only ever be triggered by compliance hooks changing the request body to something non-standard.

I am also open to a completely different implementation here, this was just the option that I saw would solve my problem and looked to be fully backwards compatible with anyone's existing `refresh_token_request` compliance hooks.

### DIFF
--- a/requests_oauthlib/compliance_fixes/__init__.py
+++ b/requests_oauthlib/compliance_fixes/__init__.py
@@ -7,3 +7,4 @@ from .mailchimp import mailchimp_compliance_fix
 from .weibo import weibo_compliance_fix
 from .plentymarkets import plentymarkets_compliance_fix
 from .ebay import ebay_compliance_fix
+from .wix import wix_compliance_fix
--- a/requests_oauthlib/compliance_fixes/wix.py
+++ b/requests_oauthlib/compliance_fixes/wix.py
@@ -1,0 +1,36 @@
+import json
+from oauthlib.common import urldecode
+
+"""
+Wix requires the request body for token requests to be sent in JSON format
+instead of x-www-form-urlencoded.
+"""
+def wix_compliance_fix(session):
+
+    def _non_compliant_access_token_request_body(
+            url: str, headers: dict, request_kwargs: dict
+    ):
+        """
+        Move the request body from the `data` kwarg to the `json` kwarg,
+        and set the `Content-Type` header to `application/json`.
+        """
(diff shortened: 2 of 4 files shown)


JoepvandenHoven-Bluemine:
Instead of relying on a ValueError would it not make more sense to inspect the Content-Type in the headers? If the value is not x-`www-form-urlencoded` then the Content-Type should be something other than `application/x-www-form-urlencoded` and conversely if the Content-Type is not `application/x-www-form-urlencoded` then it makes little sense to force the value to be `x-www-form-urlencoded`.

Relying on a ValueError is a bit iffy since it relies on the underlying oauthlib function to throw the correct Exceptions.

skray:
@violuke @JoepvandenHoven-Bluemine good points, I'm happy to change the logic to check the content-type, instead of handling the ValueError.

Before doing any more work here I'd like to make sure this PR will be accepted by one of the project maintainers, then I can do any other changes needed and update tests all at once.

skray:
I had some extra time on my hands so I went ahead and updated the code based on feedback so far.

Does anyone have advice for getting this reviewed by a maintainer?

gilou:
I don't know how to have that reviewed, 2 years later, but I agree this is an issue, and not only for Wix.

I wonder why we couldn't use the same logic as the access_token_request hook: do the dict(urldecode(body)) before calling the hook, and then provide this as an extra kwargs array, set as data by default, so that a hook can do what is required, especially send a json parameters to the request? I'll provide a PR in that spirit, however: this will break existing hooks. But that hook had an issue to begin with, as compared to the token_request one ;)

I mentioned this in #544.

PR #546, opened 2024-04-21
docs - updates Linkedin example to follow new tested pattern

jtroussard:
Updating the Linkedin example so that it follows the new pattern provided by #530


jtroussard:
@JonathanHuot might need some guidance on the scope of the unit test, also maintainer access to the vault might be more granular, however I have not tried to push anything to secrets using the cli yet.

JonathanHuot:
Wow thanks for the PR let me review.

PR #565, opened 2025-08-30
Add automatic expires_at calculation from expires_in

### LINKED ISSUE #561: you still need to update expires_in to trigger the refresh.
why!?!

injisg:
This change adds automatic calculation of expires_at from expires_in in the OAuth2 token response. This improves the automatic token refresh functionality by ensuring that expires_at is always set correctly without requiring manual intervention.

- Added automatic calculation of expires_at when expires_in is present
- Uses float precision for accurate expiration tracking
- Maintains backward compatibility with existing code
- All tests pass on Python 3.9

Fixes #561

## Documentation
The "Refreshing tokens" section in `oauth2_workflow.rst` should be updated to mention that `expires_at` is now automatically calculated from `expires_in`.

## Testing
- All tests pass locally on Python 3.9
- Added test cases for `expires_at` calculation
- Compliance fix tests updated to handle float timestamps

### DIFF
--- a/requests_oauthlib/oauth2_session.py
+++ b/requests_oauthlib/oauth2_session.py
@@ -1,5 +1,7 @@
 import logging
 import time
+import calendar
+from datetime import datetime
 
 from oauthlib.common import generate_token, urldecode
 from oauthlib.oauth2 import WebApplicationClient, InsecureTransportError
@@ -172,21 +174,37 @@ class OAuth2Session(requests.Session):
         """
         return bool(self.access_token)
 
-    def _add_expires_at(self, token):
-        """Add expires_at to token if expires_in is present and expires_at is not.
+    def _add_expires_at(self, token, response_date=None):
+        """Add expires_at to token if expires_in is present.
         
         OAuth2 responses often include expires_in (seconds until token expires) but
         oauthlib expects expires_at (timestamp when token expires) for expiration checks.
         
+        Uses response Date header if provided. Falls back to current time if
+        Date header is not provided or malformed.
+        
         :param token: OAuth2 token dict
+        :param response_date: Optional Date header value from response (e.g. "Thu, 14 Mar 2024 08:30:00 GMT")
         :return: Token dict with expires_at if expires_in was present
         """
         if token and 'expires_in' in token:
-            # Keep full precision of time.time() for accurate token expiration
             # RFC 6749 requires expires_in to be 1*DIGIT, but some providers send it as string
             expires_in = int(token['expires_in'])
-            # Keep float precision for expires_at since it's an internal field
-            token['expires_at'] = time.time() + expires_in
+            
+            # Try to use response Date header if provided
+            if response_date:
+                try:
+                    # Parse HTTP date format (RFC 7231)
+                    dt = datetime.strptime(response_date, "%a, %d %b %Y %H:%M:%S GMT")
+                    # Convert UTC time tuple to Unix timestamp (returns integer)
+                    token['expires_at'] = calendar.timegm(dt.utctimetuple()) + expires_in
+                    return token
+                except (TypeError, ValueError):
+                    # Skip if Date header is malformed or uses non-standard format
+                    log.debug("Failed to parse Date header: %s", response_date)
+            
+            # Fall back to current time (truncate to second for conservative expiry)
+            token['expires_at'] = int(time.time()) + expires_in
         return token
 
     def authorization_url(self, url, state=None, **kwargs):
@@ -422,7 +440,7 @@ class OAuth2Session(requests.Session):
             r = hook(r)
 
         self._client.parse_request_body_response(r.text, scope=self.scope)
-        self.token = self._add_expires_at(self._client.token)
+        self.token = self._add_expires_at(self._client.token, response_date=r.headers.get('Date'))
         log.debug("Obtained token %s.", self.token)
         return self.token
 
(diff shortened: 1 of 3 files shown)


requests_oauthlib/oauth2_session.py
```diff
@@ -404,7 +422,7 @@ def fetch_token(
             r = hook(r)
 
         self._client.parse_request_body_response(r.text, scope=self.scope)
```
hramrach, line 442: Doesn't `r` here have the actual time the server sent the response?

That should be more reliable than `time.time()` considering delays and clock skew between hosts.

hramrach: Also should completely avoid fiddling with floats because both is in whole seconds.

injisg: @hramrach would that just be the `Date` header in response? 

I am not sure if we're always guaranteed to have that so wdyt about a fallback to `time.time()` if not present?

injisg: @hramrach - addressed your feedback! can you take another look?

hramrach: Indeed, HTTP does not guarantee a date header.

    A sender that generates a Date header field SHOULD generate its field value as the best available approximation of the date and time of message generation. In theory, the date ought to represent the moment just before generating the message content. In practice, a sender can generate the date value at any time during message origination.

     An origin server with a clock (as defined in [Section 5.6.7](https://httpwg.org/specs/rfc9110.html#http.date)) MUST generate a Date header field in all [2xx (Successful)](https://httpwg.org/specs/rfc9110.html#status.2xx), [3xx (Redirection)](https://httpwg.org/specs/rfc9110.html#status.3xx), and [4xx (Client Error)](https://httpwg.org/specs/rfc9110.html#status.4xx) responses, and MAY generate a Date header field in [1xx (Informational)](https://httpwg.org/specs/rfc9110.html#status.1xx) and [5xx (Server Error)](https://httpwg.org/specs/rfc9110.html#status.5xx) responses.
     
When storing the Date value using time.time() as fallback is appropriate. I don't think storing the fractional seconds is needed, or adds any value.

requests_oauthlib/oauth2_session.py
```diff
@@ -171,6 +174,39 @@ def authorized(self):
...
+        if token and 'expires_in' in token:
+            # RFC 6749 requires expires_in to be 1*DIGIT, but some providers send it as string
+            expires_in = int(token['expires_in'])
+            
+            # Try to use response Date header if provided
+            if response_date:
+                try:
+                    # Parse HTTP date format (RFC 7231)
+                    dt = datetime.strptime(response_date, "%a, %d %b %Y %H:%M:%S GMT")
+                    # Convert UTC time tuple to Unix timestamp (returns integer)
```
hramrach, line 199: The GMT here should be parsed as %Z. Otherwise the time is likely wrong.

hramrach:
Also in general this adds the `expires_at` field which is not standard and nothing will use it.

For a full fix something needs to use the new field instead of `expires_in` for the automated refresh to happen.

injisg:
@hramrach my initial thought was that the oauth library would raise a TokenExpiredError [here](https://github.com/oauthlib/oauthlib/blob/master/oauthlib/oauth2/rfc6749/clients/base.py#L215-L216) which would result in refreshing of the token [here](https://github.com/requests/requests-oauthlib/blob/f33dac32bd065b835b34e7f0a48025700aae1e31/requests_oauthlib/oauth2_session.py#L532) in requests-oauthlib. 

But I see now that oauthlib only checks _expires_at - shouldn't it check expires_in at least? and to make this work, we'd need to make it support the new expires_at field too? Let me know if that sounds right! and i can create an issue / put up a fix!

hramrach:
Indeed, although the expires_at is non-standard it is already supported. Then calculating it when not present should suffice. Thanks

requests_oauthlib/oauth2_session.py
hramrach: Does not seem to make any difference:

```
TZ=PDT python3.13
Python 3.13.5 (main, Jun 12 2025, 00:40:24) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from datetime import datetime
>>> datetime.strptime('Tue, 02 Sep 2025 10:28:50 GMT', "%a, %d %b %Y %H:%M:%S GMT")
datetime.datetime(2025, 9, 2, 10, 28, 50)
>>> datetime.strptime('Tue, 02 Sep 2025 10:28:50 GMT', "%a, %d %b %Y %H:%M:%S %Z")
datetime.datetime(2025, 9, 2, 10, 28, 50)
>>> datetime.strptime('Tue, 02 Sep 2025 10:28:50 GMT', "%a, %d %b %Y %H:%M:%S %Z").tzinfo
>>> datetime.strptime('Tue, 02 Sep 2025 10:28:50 GMT', "%a, %d %b %Y %H:%M:%S GMT").tzinfo
```

See also https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow


>Warning
>
> Because naive datetime objects are treated by many datetime methods as local times, it is preferred to use aware datetimes to represent times in UTC. As such, the recommended way to create an object representing the current time in UTC is by calling datetime.now(timezone.utc).

injisg: i'll still parse with %Z for consistency

hramrach: I would hope that parsing with %Z would make `strptime` generate the `tzinfo` but no, there is no `tzinfo` either way.

As the warning in the documentation suggests handling of time data in python is less than ideal.

hramrach: To spell it out more clearly: with timezones python leans towards implicit rather than explicit, and in the long run that is a mess because you never know in what timezone your timestamp is.

Then to get the best possible result you would parse GMT as fixed string. Then you know that the timastamp is GMT, and can manually add explicit `tzinfo` hardcoded to `timezone.utc` making the timestamp unambiguous.

Or you can forget all this, leave the code as is, hope it works with the implicit timezones.

